Issue #22, opened 2014-10-06
Missing single quote in Cache Remote Requests Example


Faison:
Hi there,

Wasn't sure if me forking this repo would create a private fork or not, and this issue is so small that I decided it wouldn't hurt to just bring it up to someone who's already committing.

Basically, if you look at the code example in the section titled [Cache Remote Requests](http://10up.github.io/Engineering-Best-Practices/php/#cache-remote-requests), there's a line that looks like this:

`wp_cache_set( 'prefix_other_blog_posts, $posts, '', HOUR_IN_SECONDS );`

There's a closing single quote missing right after `prefix_other_blog_posts`.

Like I said, really small. If it is safe for me to fork this repo without exposing it to the world, let me know and I can send a pull request, otherwise this issue is in the file _includes/markdown/PHP.md on line 220.

Thanks,
Faison

rachelbaker:
Fix in PR #23

PR #23, opened 2014-10-07
Syntax and Standards fixes for PHP Code Examples

rachelbaker:


DrewAPicture:
I'd like to see docblocks added to our example functions as well. I'm happy to do it if there's agreement.

rachelbaker:
@DrewAPicture YES, please.  But get your own issue for that, and don't bogart mine :kissing_cat:

DrewAPicture:
Seeing as docs fall under standards, I'd say it applies. But I'm picking up what you're laying down :)

rachelbaker:
@DrewAPicture I did open the door by going outside of #22's narrow scope and adding in spacing adjustments.  You are correct, that docs fall under standards.  

This PR is sourced from my fork.  Which means that really the best course of action here is to have a new issue (or PR) specific to the code standardization of our examples.

tlovett1:
Looks good. Agree, separate PR.

PR #24, opened 2014-10-07
Add phpDoc blocks to example PHP functions.

DrewAPicture:
Also various inline-comment syntax adjustments.


tlovett1:
Looks great. Can you fix the conflict?

DrewAPicture:
There you go.

PR #32, opened 2014-11-05
fix mixed mode asset requests.

paulirish:
Because right now it looks like this:

image

and these are the blocked requests:

image


rachelbaker:
@paulirish  Good catch and thank you for the PR :dancers:

Issue #38, opened 2014-11-06
Confusing WP_Query vs. get_posts()


JDGrimes:
On the PHP page under the [Effecient Database Queries section](https://github.com/10up/Engineering-Best-Practices/blob/gh-pages/_includes/markdown/PHP.md#efficient-database-queries) is this:

> WP_Query actually calls get_posts(); calling get_posts() directly bypasses a number of filters. Not sure whether you need these things or not? You probably don't.

I find that very confusing. It is actually the other way around: [the `get_posts()` function uses `WP_Query`](https://core.trac.wordpress.org/browser/tags/4.0/src/wp-includes/post.php#L1866). I assume that maybe you meant to refer to the _`WP_Query::get_posts()` method_.

tlovett1:
You are correct. Thank you.

PR #39, opened 2014-11-06
WP_Query doesn't call get_posts()

kovshenin:
`WP_Query` doesn't call `get_posts()`, it's the other way round. `get_posts()` creates a new `WP_Query` object and indeed defaults `suppress_filters` to `true`. There's a gotcha too. If you need a list of posts in a widget, `WP_Query` will prepend sticky posts by default, which may or may not be intended behavior, while `get_posts()` defaults to `ignore_sticky_posts`.


tlovett1:
Thanks @kovshenin. Closes #38

PR #40, opened 2014-11-06
Gruntfile.js is the name of the grunt file

aaronjorbin:
This allows the line describing Gruntfile.js to follow the same standard as the others in the list where the name of the file is in code.


rachelbaker:
:cake: 

Thanks @aaronjorbin

PR #44, opened 2014-11-11
it's/its fixes

trepmal:


trepmal:
Pronouns like to be difficult and are the exception to the "apostrophe for possession" rule

tlovett1:
Interesting. Learned something new :)

PR #48, opened 2014-11-12
wp_nonce_field instead of wp_create_nonce

brandondove:
If you want to output a nonce field, you're using the wrong function.


tlovett1:
Nice catch.

PR #49, opened 2014-11-12
FIxed broken link

thinkstylestudio:
Package manager link was pointing back to version control page insead of tools

`[package managers](#package-managers)` to

`[package managers](/Engineering-Best-Practices/tools/#package-managers)`


rachelbaker:
@thinkstylestudio Good catch on the broken link.  To correct the link, it would be best to use the `{{ HOME_PATH }}` variable instead of hard-coding the relative path from the `/Engineering-Best-Practices/` segment.  

Can you update your PR to use `{{ HOME_PATH }}/tools/#package-managers` ?

thinkstylestudio:
Sweet, figured it out. Let me know if that looks good.

PR #50, opened 2014-11-19
Docblock Long Description Emphasis

jamesmehorter:
Hey all.. trying to put some more emphasis on the use of verbose docblock long descriptions to communicate code purpose, intent, and context. Ideally, each function/heavy logic section of code (any language!) should include a long description primarily for future maintainability. Most assuredly, another engineer will someday need to work on your code, when you're long gone. Knowing WHY the code was written, and what it's purpose is within the larger project/context is invaluable when another engineer is onboarded to make edits later on. 

The secondary purpose for these is so managers/clients can easily understand what's going on without reading the code. Think of these as code manuals, just as Ford ships one with each automobile; we should be doing the same for each piece of code we write. 

The current WordPress Documentation Standards just state that a long description should be included, but I think we can do A LOT better than that! We can emphasize documenting code on a thoughtful level, imagining the questions the next engineer would ask and answering them ahead of time. 

Please let me/us know what you think! Feedback is greatly encouraged!!! :)


GhostToast:
I currently do this as much as possible

jonbellah:
+1

rachelbaker:
@jamesmehorter Custom Fields are now hidden by default in the Edit Post/Page screen settings, so protecting meta simply because you have a meta box is not the suitable of intended use of protection.  The expected behavior when electing to view the Custom Fields of a Post or Page is that all the meta fields that are publicly viewable (and would be exposed in the JSON REST API) are listed.  

I think the wording in your Docblock just needs to be tweaked to reflect the correct practice of:
- a meta field that holds **internal read only** data = protected, via `_` prefix or the `is_protected_meta` filter
- a meta field that is intended to be a viewable component of the post (Examples: event date, or employee title) should **not** be protected.

jamesmehorter:
Thank you very much, @rachelbaker!!! I've since adjusted the language used for the example function to properly speak to the difference between public/private. 

Also made a few other little corrections/adjustments to presentation.

tlovett1:
@DrewAPicture thoughts on this?

DrewAPicture:
I agree with the intent of making the description field more verbose, though not necessarily for the express purpose of making it easy for a non-technical person to grok. If they're in the code, there would already be an expectation of knowledge to understand what's going on anyway, so simplifying this one section of a DocBlock alone isn't necessarily going to facilitate that.

I think if that's the ultimate goal, then there's a compelling argument for creating user-facing documentation that can be equally grokked by technical and non-technical folk alike.

As for the sample code + docs, I don't know that this method of marking things as "protected" is something that we want to promote – it sidesteps the process of registering meta and handling whether it's considered protected via the auth callback.

tlovett1:
Merging this. If anyone wants to amend language on meta, feel free to submit another PR. Thanks @jamesmehorter !

Issue #30, opened 2014-11-05
Add some more performance-related args to WP_Query section


helen:
We talk about `no_found_rows`, but in reality that only removes one query out of the **five** that `WP_Query` runs in its typical configuration. We should also document: `update_post_term_cache`, `update_post_meta_cache`, and `'fields' => 'ids'`.

rachelbaker:
:+1:

sc0ttkclark:
`'fields' => 'ids'` is my favorite :)

tlovett1:
@helenhousandi mind writing a PR?

ericmann:
Already been merged.

PR #54, opened 2014-12-18
More performance-related args for WP_Query section

helen:
See #30 - I am not loving the current styling, though. The subsections run together, and trying to use another list within that subsection messes up the visual hierarchy.

screen shot 2014-12-18 at 4 22 20 pm


tlovett1:
@helenhousandi Thanks! I cleaned up the markdown for this in b1e0a910

PR #57, opened 2015-01-20
Use the DOM API to Insert HTML

dana-ross:
We had feedback from a VIP code review about this last week, so I thought it would be a good topic for our Best Practices. See https://gist.github.com/daveross/d5e15b6b0ebd44e09962 for a trivial example of this issue.

This might need some editing to make it clearer to someone who isn't used to working with DOMElement objects directly. Feedback is welcome before merging.


jonathantneal:
Nice work! I like the practices outlined in this pull, and I’d like to see more. Purposeful DOM manipulation improves the dependability, longevity, and readability of JavaScript. We could outline other methods too — `classList.add` for example.

I’d love to see what VIP has to say about this, and our admins, but I would remove language that suggests these practices also provide a security benefit. What did VIP say to us about this? Otherwise, pardon my Opinions™, but here’s mine on the matter:

`setAttribute` is no more superior to `innerHTML` if a third-party library can just as easily overwrite the `setAttribute` method altogether. In contrast, we’re able to sanitize strings in PHP because those strings do not execute. Unfortunately, we are not afforded such luxuries by third-party scripts — they more closely resemble `include` statements.

Anyway, I like the example you’ve given — a situation where something is intended to be a literal attribute and instead closes out the element — because it’s a totally valid example from a dependability and readability perspective, and I remember making the mistake myself (I hope I’ve learned better). It’s important to see how badly we can mess up the DOM when we start putting together components by writing innerHTML interspersed with partials.

Thoughts? Agree? Disagree? Able to share VIP’s feedback?

dana-ross:
My only concern with classList.add() is that it isn't supported in IE 8 & 9. We still have projects where IE 8 support is expected.

Here's the general advice from VIP's feedback:

> …scripts need to be refactored to remove all instances of where HTML is created by concatenating strings. The problem with this approach is that when building HTML as strings, there aren't any restrictions on what data can be added to the DOM, which can lead to Cross-Site Scripting vulnerabilities.
> 
> The preferred approach is to build a node and then apply it to the DOM. This is safer because when setting an attribute of a DOM node, you can’t break out of the attribute. Additionally, this is also the correct layer of abstraction for dealing with the DOM."

It's true third-party scripts could override Element.prototype.setAttribute(), but I'd say that's out of our control. This is something we can control and take responsibility for.

dana-ross:
I should also add language about the [performance benefits](http://jsperf.com/native-appendchild-vs-jquery-append/12) of manipulating the DOM directly.

thaicloud:
While not a perfect solution, I would say that setAttribute does provide security benefits over innerHtml.

I would liken it to the difference between locking your front door or not before you go to bed. Sure, you could argue that I seasoned burglar can easily pick the lock if they wanted; but locking it still ensures that not just anyone can waltz through the front door.

Likewise, setAttribute does add another layer of protection from malicious users; a form input field that slips past our validation and into an innerHtml is going to run (as in Dave's example)- if it goes through setAttribute then it won't.

jonathantneal:
Good points, nice to meet you both, and thanks for sharing VIPs notes. I’ll try not to present much disagreement if we’re going to agree on the end result. Like, if the example is a burgler, IMHO you’ve already invited them into your house, given them a map of what’s in there, and entrusted them to do whatever they want as your identity. Could giving someone hope that `setAttribute` protects them do more damage than good? Am I seeming super unreasonable?

I’m also willing to contribute to this if you want to divide up work. I’d love to see notes for `document.createTextNode`, and yes `classList.add`, but also footnotes to caniuse tables. I would rather see good code and compatibility footnotes than omissions. For similar reasons, I wouldn’t want to omit `matchMedia`.

tlovett1:
Loving this thread. Thanks Dave!

I agree with Jonathan in that what we are promoting here purposeful DOM manipulation using the DOM API. This leaves less room for error (but is not foolproof to Jonathan's point).

A couple things I would like in this before merging:
1. Follow JS coding standards (proper spacing)
2. Follow existing JS best practices `jQuery('#test').html(...)` should be `jQuery( document.getElementById( 'test' ) ).html( ... )`
3. Let's remove this jQuery all together: `jQuery('#test').get(0).appendChild(new_div);` should be `document.getElementById( 'test' ).appendChild( new_div )`

dana-ross:
I wanted to keep the jQuery calls simple, but I agree we should be good role models. Instead of `jQuery( document.getElementById( 'test' ) ).html( ... )` I think I'll cache that jQuery object to a variable first so people can focus on where we're doing the actual DOM manipulation.

Thanks for the feedback everyone!

lkwdwrd:
I would love to see better clarification of the elements involved with this pull request. 

jQuery maps down to basically `el.innerHTML = string;`. However, the important component here is that it is also running the string through eval() which causes the JS to be executed. To say jQuery should be avoided for DOM work is missing the point. Understanding jQuery's behavior is what should be encouraged.

There are jQuery methods for doing element insertion too that would not map to the innerHTML property, but to the DOM methods you are talking about. You can also use jQuery to build up fragments before inserting them. I would not want to give the impression that jQuery is evil and native DOM makes things safe, because that is not true. It's a tool for normalizing browser inconsistencies and can be used well or poorly.

``` javascript
// Assume hacked_string started off as something completely fine, but a third-party script changed its value so now it looks like...
var hacked_string = "\"\>\<script\>alert(\"haxx0r3d\")\<\/script\>";

// Proper use of fragments in DOM API
var newDiv = document.createElement( 'div' );
newDiv.setAttribute( 'id', hackedString );
document.getElementById( 'testDom' ).appendChild( newDiv );

// Proper use of fragments in jQuery
var $newDiv = $('<div />');
$newDiv.attr( 'id', hackedString );
$( '#testDom' ).append( $newDiv );

// Inject content using DOM API and concatinated strings
// (not so good. doesn't execute the script, but does break the div container)
document.getElementById( 'testDom' ).innerHTML = '<div id="' + hackedString + '">content</div>';

// Inject content using jQuery and concatinated strings.
// (bad because it breaks the div container and runs through eval() which executes the script).
$('#testDom').html('<div id="' + hackedString + '">content</div>');
```

There are times when using `.html()` or `innerHTML =` is necessary. Using [JS templating](https://lkwdwrd.com/wp-template-js-templates-wp/) for example. JS templates at this point get into the DOM through `.html()` or `.innerHTML =` style injection. However in a templating library worth its salt, there will be a way to escape strings before adding them into the template for insertion.

There are sometimes clear reasons for choosing to use jQuery or DOM API. DOM is faster because jQuery selection/manipulation is an abstraction/normalization of the DOM API. I'm all for avoiding the sizzle selection engine when browser support permits (almost always these days). I'm all for ditching jQuery when browser support/team comfort permits (woot, 30k less scripts to add to the page). However, the reasons are almost always performance based, not security based. The security piece lies in better understanding the tool.

ericmann:
The issue here isn't so much that we need to call out direct DOM manipulation as a best practice, but that we need more emphasis on properly sanitizing inputs. Whether working with native JavaScript, jQuery, or even PHP on the server side, we _never_ trust client input because it could be malicious.

Put another way, you never concatenate strings in PHP with untrusted `$_GET` or `$_POST` values.  Instead, we _always_ sanitize this data first - just like we _always_ escape dynamic data before printing it to markup.

JavaScript is no different. Passing variables around is great, but we need to take care to sanitize those variables before we use them and to escape those variables before we print things to the page. There isn't a JavaScript equivalent of `sanitize_text_string()` or `esc_html()` yet, but that's the kind of mindset we need here.

To be clear, there are currently some minor issues within _jQuery Migrate_ that cause this to present as an issue. I've raised them with the appropriate security contacts at VIP, WordPress.org, and even jQuery core - teams on all sides are working on fixing the issues right now. That said, the issues can be avoided by focusing not on using the DOM but on properly sanitizing inputs and escaping outputs.

In other words, there is no problem with `$( '<div id="my_element"></div>' )`. It's functionally succinct, maintainable, and readily understood by anyone looking at the code in question. However, anything with dynamically-concatenated values is suspect as those variables are not necessarily escaped. `$( '<div id="' + some_var + '"></div>' )` could pass _anything_ into the `some_var` field  and is insecure for the same reason `"SELECT * FROM wp_posts WHERE post_title='" . $title . "';"` is insecure.

I'm all for strengthening our best practices when it comes to an understanding of security and how we pass data in and out of the various languages, constructs, and libraries we use. At the same time, we need to be very careful about confusing security practices with performance practices or code styling guidance. The DOM is powerful, and using direct DOM manipulation in the linked example will plug the potential security hole. That said, making DOM a standard as a result doesn't address the root issue or put us on any safer ground.

helen:
jQuery's `.text()` (which I believe is really calling `.createTextNode()`) has an end result that is not unlike `esc_html()`, if that is worth mentioning at all. I see this come up more frequently than I'd like.

lkwdwrd:
[Underscore JS's Escape method](http://underscorejs.org/#escape) gets us a little ways down the road with regards to HTML escaping as well, but it's not always present. There are definitely ways to skin this cat.

jonathantneal:
Eric is talking about sanitizing user input, wheras before we were talking about third-party libraries. To clarify:

Sanitizing user input is like writing a chat application that filters out HTML injections. 

``` js
input.value = '<alert>chat.send([document.cookie, JSON.stringify(localStorage), navigator.userAgent]).then(function () { alert("Virus detected in Real Media Player"); }); </alert>';
```

Loading third-party libraries is like taking your vows with another website.

``` js
document.head.appendChild(document.createElement('script')).src = 'http://youcan.trustme.baby/u-the-real-mvplayer.js'; // I trust you so bad
```

ericmann:
@jonathantneal By "user input" I mean any untrusted data source. In the case of jQuery and the known issue of the hole in jQuery Migrate, a user could be an input field in a page, it could be data returned by an ajax call, it could be a script injected by a third party library, or it could be a parameter passed into [the URL hash](https://eamann.com/tech/jquery-xss/).

jonathantneal:
@daveross, the most recent versions are super awesome! All hail vanilla JavaScript!

@ericmann, would it be fair to say one should **only** load third-party libraries from trusted sources? I’m in agreement with most of the sources you’ve specified, but third-party libraries are not like JSON or input.value or location.hash. In PHP terms, they’re like an even more powerful version of `include`. I wouldn’t want to cut it any other way, even if muddying things with something like JSONP.

ericmann:
@jonathantneal I guess I'm using the wrong terms. By "source" I'm specifically referring a source of data. `$_GET` is a source of data in PHP. `window.location.hash` is a source of data in JS. Effectively, `$.html()` is a _sink_ in code that assumes all data passing into it is sanitized already. Passing data from an untrusted _source_ to a trusted _sink_ is the issue we're facing. If the data starts as untrusted, then it needs sanitization, otherwise the transfer to a trusted sink is a potential security violation.

All of that said, yes, we should only be loading libraries from trusted sources. Specifically, we should only be loading libraries from within our own distribution. When we load libraries from _external_ sources they should be a) fully vetted, b) fully documented, and c) as isolated from our own code as possible. This is the entire reason we wrap our JS code in a closure, passing a global `this` in as `window` so we get unadulterated copies of all of the window-derived objects and methods we expect to have available.

At the end of the day, we're having two conversations here. The first, about the use of DOM methods for HTML manipulation of jQuery (specifically because of the issue I reported to VIP that they in turn passed along in a code review). The second is about the use of third-party libraries that may or many or may not rewrite the JS API in ways we don't expect. Both are important conversations, but the second isn't relevant to the pull request.

_includes/markdown/JavaScript.md
```diff
@@ -164,6 +164,50 @@ jQuery( '.menu' ).on( 'click', function() {
 
 Another example in JavaScript is ```escape()``` and ```unescape()```. These functions were deprecated. Instead we should use ```encodeURI()```, ```encodeURIComponent()```, ```decodeURI()```, and ```decodeURIComponent()```.
 
+<h3 id="dom-api">Use the DOM API to Insert HTML {% include Util/top %}</h3>
+
+When adding elements to the page in JavaScript, create new DOMElement objects using document.createElement(), then set their properties using setAttribute() and the className property. This ensures that the attribute values you set are used literally, even if they contain markup. jQuery's html() function doesn't perform any sanitization on the string passed to it, so it's not secure to build your HTML as a string and inject it into the page this way.
```
tlovett1: Agree with Eric/Luke, there needs to be some clarification here that jQuery.html() is not insecure but can be easily misunderstood and passed unsanitized data.

NicoleGeissinger:
If you have access to the WordPress VIP lobby, Nick Daugherty wrote up a great post around this very subject last Friday. Just thought I'd throw that on the radar. http://lobby.vip.wordpress.com/2015/02/06/preventing-xss-in-javascript/

I believe to get access you need a VIP approved account and will need to request access to it.

dana-ross:
This PR is pretty stale. I'm going to close it out.

PR #58, opened 2015-01-22
WP_Query vs. get_posts() vs query_posts()

Jaace:
New section added outlining differences between WordPress querying


tlovett1:
Thanks Jason! I'm going to merge and do some slight rephrasing of a few things.

PR #66, opened 2015-04-29
tweaked some language, grammar, and added a main tag to a markup example

andrewmowe:


tlovett1:
@JonBellah thoughts on this?

jonbellah:
I agree with all of the changes in this PR. I especially like the clarity brought to the Progressive Enhancement section, that was one that I was having particular difficulty with.

PR #67, opened 2015-04-30
Fixing FOUC issue that occurs when clicking links found in the table of ...

jonbellah:
...contents, which results, oddly enough, from no slash being present between the page URL and the anchor tag


tddewey:
Super weird. Thanks for tracking it down!

PR #71, opened 2015-05-04
CSS: Add a blank line before the inheritance code block.

ocean90:
See the following screenshot from https://10up.github.io/Engineering-Best-Practices/css/#inheritance

CSS Inheritance


lkwdwrd:
Good catch, thanks!

PR #76, opened 2015-05-14
Made readability improvements to content and code blocks at mobile widths...

jonbellah:
Also, repositioned the navigation menu so that it does not wrap or overlap as the browser shrinks - this also brings our Best Practices site in line with the design on 10up.com

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -17,6 +17,6 @@ redcarpet:
 sass:
   sass_dir: _includes/_sass
 CONFIG:
-  BASE_PATH: https://10up.github.io/Engineering-Best-Practices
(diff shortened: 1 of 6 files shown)


tlovett1:
The code blocks still appear "crunched" in narrow viewports. Was this intended to fix that?

jonbellah:
Hmm, can you provide a screenshot? I've tested in a number of browsers and virtual devices. I've also tested on my physical iPhone 6plus... Attached screenshots of that test. If you're seeing issues locally, it may be because you haven't updated the BASE_PATH to reference your local build.
first
two

tlovett1:
Screenshot for code crunching. This is from PHP performance:

10up_engineering_best_practices

PR #78, opened 2015-05-14
Feature/mobile adjustments

saltcod:
It occurred to me after I opened that h3 style ticket that most of what I've done today is pretty subjective. So here's the rest of it. Mostly addresses spacing / padding / margin issues on mobile, but also makes a bunch of font adjustments. 

I'm new, new, new to this so 1000 apologies for likely doing this the wrong way!   =)

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -17,6 +17,6 @@ redcarpet:
 sass:
   sass_dir: _includes/_sass
 CONFIG:
-  BASE_PATH: https://10up.github.io/Engineering-Best-Practices
+  BASE_PATH: http://127.0.0.1:4000
 
   ASSET_PATH: false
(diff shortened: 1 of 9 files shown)


saltcod:
Hummmm so ....I can keep adding to this pull request?  I probably shouldn't though (?). 

Anyway, the last commit here addresses an issue with the menu at in-between window sizes. As it is now, the toggled menu needs to be invoked earlier to avoid the main nav menu jumping to the next line as per the screen. Another solution would be to progressively drop the last n items as the screen gets narrower. I like this approach sometimes. 

hamburgler

jonbellah:
The issue in your screenshot is addressed in https://github.com/10up/Engineering-Best-Practices/pull/76

tlovett1:
Can you guys coordinate on these PRs?

tlovett1:
Any update on this?

tlovett1:
Closing. Feel free to reopen if still relevant.

PR #82, opened 2015-06-23
Integrations

ericmann:
Add an Integrations section to project structure. Includes:
- How to document 3rd party integrations
- How to prevent production integrations from firing in development
- Strong caution against ever hard-coding production credentials anywhere

### DIFF
--- a/_includes/markdown/Structure.md
+++ b/_includes/markdown/Structure.md
@@ -18,7 +18,7 @@ Remote service for fetching funding and other investment data related to tech st
 - /includes/classes/cron.php
 
 ### Development API
-- u/p: devuser/adfdlewrkw123
+- See http://somesitethatrequireslogin.com/credentials-for-project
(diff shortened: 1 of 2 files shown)


morganestes:
I'm curious as to why it's encouraged to put development API keys/credentials in a text file, instead of storing them outside of the repo completely and referencing that location in the Integrations file? It seems like having _any_ credentials in a file that could get leaked out is a bad idea.

ericmann:
@morganestes I can live with that. For most dev-side APIs and private repositories, it's not too huge of an issue. The idea here is to protect from _production_ keys ever being used in development/staging and leading to an embarrassing release of information.

That said, I'm OK just saying "no keys in repositories" to prevent dev accounts from being accidentally leaked down the road.

tlovett1:
No keys in repos sounds good to me.

PR #85, opened 2015-06-24
use isset instead of in_array

nicoladj77:
Adding a section to PHP performance page to explain why isset() should be used instead of in_array()

### DIFF
--- a/_includes/markdown/PHP.md
+++ b/_includes/markdown/PHP.md
@@ -87,7 +87,7 @@ As outlined above, `get_posts()` and `WP_Query`, apart from some slight nuances,
 As noted in the [WordPress Codex (along with a useful query flow chart)](http://codex.wordpress.org/Function_Reference/query_posts), `query_posts()` isn't meant to be used by plugins or themes. Due to replacing and possibly re-running the main query, `query_posts()` is not performant and certainly not an acceptable way of changing the main query.
 
 ##### Use isset() instead of in_array()
-`isset()` should always be used instead of in_array() as
+[`isset()`](http://php.net/manual/it/function.isset.php) should always be used instead of [`in_array()`](http://php.net/manual/it/function.in-array.php) as
 
  * It uses an `O(1)` hash search on the key whereas `in_array` must check every value until it finds a match, thus its complexity is `O(n)`
  * Being an [opcode](http://3v4l.org/pIuZV/vld#tabs), it has less overhead than [calling the in_array](http://3v4l.org/S4YOA/vld#tabs) built-in function.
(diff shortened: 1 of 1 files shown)


scottlee:
Any update here, @tlovett1?

dana-ross:
I think the big-O notations should have links out to external pages explaining what they mean, for engineers without a computer science background.

Also, isset() considers something to not be set if its value is null. This may be an edge case, but we should note this exception to the rule.

tlovett1:
I may be confused but would `in_array` even work in the example since you are searching the array keys? Also, I would assume `array_flip` is O(n) so doing `array_flip` plus `isset` is O(n + 1). I get that `isset` is an opcode but that performance gain seems pretty minor to me.

dana-ross:
Agreed this should be `array_key_exists()`. `in_array()` has nasty undocumented behavior when not in strict mode that sometimes has it return `true` for array keys _and_ for keys/values that don't even exist.

nicoladj77:
@tlovett1 maybe for clarity i should add that this means creating arrays differently  from what normally it's used. normally you would do

``` php
$array = array(  'foo', 'bar' );
if ( in_array( 'foo', $array ) ) {
  // value is present in the array
};
```

i'm suggesting to change the way the array is created in the first place so that you can use `isset()` like in the example. 
[Here are some benchmarks](http://maettig.com/1397246220) of isset vs array_key_exist vs in_array. 
I would say that `isset` is still better than `array_key_exist` as it's a language construct instead of a function, even with the null caveat, as usually null in php is [used to unset values](http://php.net/manual/en/language.types.null.php), but both of them are equally good.
I added the array_flip part only because over the internet there [are suggestions](http://php.net/manual/en/function.array-flip.php#106895) that it might be a perfect solution for every case, which is not. We can remove that part if it's confusing, I just meant that it "might" be useful in limited occasions but it would be better to benchmark first
one of the reasons that  prompted me to write this is the fact that WordPress VIP has started signalling all in_array calls, asking them to be removed even if removing them would mean rewriting arrays to have a different structure. 

> "in the xxx function you are doing a lot of in_array() checks on a large size array, the speed will be O(n) if you change to looking up the values in a hash table (using the array key to lookup and not array value) you will get O(1) which is much better in arrays that contain many elements."

so i guess this is something that should be addressed in our best practices

dana-ross:
@tlovett1 Yes, `array_flip` is O(n). It uses ZEND_HASH_FOREACH_KEY_VAL

dana-ross:
Ultimately, I feel this is bigger than `isset` vs `array_key_exists`. VIP's feedback is correct, we shouldn't be using an O(n) operation to search for things in arrays when there are O(1) methods available. Yes, this may require us to structure arrays differently in the first place, but that's part of the job.

I'd like to see this re-written from that perspective, making it about building arrays that encourage lookups by key rather than searches by value.

Within that topic, we can add a note that `isset($array['key'])` is faster and explain why. But frankly that's a micro-optimization compared to the O(n) elephant in the room.

nicoladj77:
@daveross sure, that's exactly what i wanted to say! It wasn't an `isset` vs `array_key_exists` discussion/flame :smile: 
I'll rewrite this with the perspective you suggested, and adding links for Big O notations.

nicoladj77:
@daveross updated!

Issue #90, opened 2015-07-15
Specific guidance on namespaces


dana-ross:
We've discovered inconsistencies with how namespaces are applied on some projects. Now that 10up engineers are adopting namespaces more widely in their projects, we should be providing more specific guidelines on how to use them in our projects.

A pull request will be coming after some thought on my end, but I wanted to open an issue to get my thoughts together. Any input is welcome.

> We do this to logically organize our code and to prevent collisions in the global namespace

What does this mean to someone with no experience with namespaces? What's the "global namespace"? How do things "collide" there? Was anybody hurt?

Let's explain this more clearly, perhaps with a link to an external resource with a good explanation for beginners.

> namespace tenup\Utilities\API;

We need to set a standard namespace for 10up projects, and include general guidance for non-10up developers who want to learn from our docs. I propose we stick with this "tenup" namespace that's already in our best practices, although maybe changed to "Tenup" or "TenUp" for consistency with PSR-compliant projects out there.

From there, I propose the next level of the namespace be the client's name (i.e. "tenup\Tyrell-Corporation") followed by a particular site or product name ("tenup\Tyrell-Corporation\Nexus-6"). Consistent namespace naming on a project helps autocompletion in IDEs like phpStorm, making development more efficient, making engineers more productive and EMs happier. On projects where there is a "shared plugin" with common functionality for different sites, we can call that sub-namespace "Util" or "Common".

Theme code and plugins within a project can share a namespace. If there's a collision inside that namespace, it's a sign we probably have redundant code, not a prompt to create a new namespace. Also, adding more namespaces means more "use" declarations or long \paths\to\namespaces\in_function_calls() which makes code harder to read.

dana-ross:
> All functional code should be properly namespaced.

This most likely doesn't refer to _functional programming_ and should be changed to "procedural code", "non-OOP code", or re-phrased completely to keep engineers from getting confused when they google for help.

(maybe "non-template code" is closer to what's meant here?)

scottlee:
Totally +1 this idea, Dave. The docs as they stand now don't provide much insight into namespacing beyond the recommendation. I too would like to see further clarification around our internal naming policies for what comes after tenup. And, fwiw, I'm all for moving to TenUp vs. tenup, but only because it looks nicer.

tlovett1:
[...]

PR #91, opened 2015-07-15
First stab at more guidance regarding namespaces

dana-ross:
Related to issue #90

### DIFF
--- a/_includes/markdown/PHP.md
+++ b/_includes/markdown/PHP.md
@@ -363,11 +363,15 @@ function do_something() {
 }
 ```
 
-A namespace identifier consists of a _top-level_ namespace or "Vendor Name", which is usually ```TenUp``` for our projects. We follow the top-level name with a project name, usually a client's name, and then with the name of a particular site or high-level project we're building. Themes and plugins written for the same client should share a namespace. If we have two identically-name classes or functions in the same namespace, that's a sign there might be duplicate code we can get rid of.
+A namespace identifier consists of a _top-level_ namespace or "Vendor Name", which is usually ```TenUp``` for our projects. We follow the top-level name with a project name, usually a client's name. ex: ```TenUp\Buy_N_Large;```
   
-When we manage multiple sites for a single client and create a _shared plugin_ for common functionality, it should have its own namespace named "Common" (example: TenUp\Tyrell-Corporation\Common").
+Additional levels of namespace are defined at discretion of the project's lead engineers. Around the time of a project's kickoff, they agree on a strategy for namespacing the project's code. For example, the client's name may be followed with the name of a particular site or high-level project we're building (```TenUp\Buy_N_Large\Wall_E;```).
 
-[```use``` declarations](http://php.net/manual/en/language.namespaces.importing.php) should be used for classes outside a file's namespace. By declaring the full namespace of a class we want to use *once* at the top of the file, we can refer to it by just its class name, making code easier to read.
+When 10up works on more than one project for a client and we build common plugins shared between sites, "Common" might be used in place of the project name to signal this code's relationship to the rest of the codebase.
+
+The engineering leads document this strategy so it can be shared with engineers brought onto the project throughout its lifecycle.
+  
+[```use``` declarations](http://php.net/manual/en/language.namespaces.importing.php) should be used for classes outside a file's namespace. By declaring the full namespace of a class we want to use *once* at the top of the file, we can refer to it by just its class name, making code easier to read. It also documents a file's dependencies for future developers.
 
 ```php
 /**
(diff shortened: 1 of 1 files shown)


_includes/markdown/PHP.md
```diff
@@ -350,17 +350,41 @@ Using a common set of design patterns while working with PHP code is the easiest
 
 #### Namespacing
 
-All functional code should be properly namespaced. We do this to logically organize our code and to prevent collisions in the global namespace. Generally, this means using a PHP ```namespace``` identifier at the top of included files:
+We should properly namespace all PHP code outside of theme templates. This means any PHP file that isn't part of the [WordPress Template Hierarchy](https://developer.wordpress.org/themes/basics/template-hierarchy/) should be organized within a namespace or _pseudo_ namespace so its contents don't conflict with other, similarly-named classes and functions ("namespace collisions").
```
scottlee: Should we drop the "should" to be more precise?

_includes/markdown/PHP.md
```diff
@@ -350,17 +350,41 @@ Using a common set of design patterns while working with PHP code is the easiest
...
 function do_something() {
   // ...
 }
 ```
 
+A namespace identifier consists of a _top-level_ namespace or "Vendor Name", which is usually ```TenUp``` for our projects. We follow the top-level name with a project name, usually a client's name, and then with the name of a particular site or high-level project we're building. Themes and plugins written for the same client should share a namespace. If we have two identically-name classes or functions in the same namespace, that's a sign there might be duplicate code we can get rid of.
+  
+When we manage multiple sites for a single client and create a _shared plugin_ for common functionality, it should have its own namespace named "Common" (example: TenUp\Tyrell-Corporation\Common").
+
+[```use``` declarations](http://php.net/manual/en/language.namespaces.importing.php) should be used for classes outside a file's namespace. By declaring the full namespace of a class we want to use *once* at the top of the file, we can refer to it by just its class name, making code easier to read.
```
scottlee: Great idea.

_includes/markdown/PHP.md
dana-ross: Yes.

_includes/markdown/PHP.md
```diff
@@ -350,17 +350,41 @@ Using a common set of design patterns while working with PHP code is the easiest
...
+namespace TenUp\Tyrell-Corporation\Nexus-6;
 
 function do_something() {
   // ...
 }
 ```
 
+A namespace identifier consists of a _top-level_ namespace or "Vendor Name", which is usually ```TenUp``` for our projects. We follow the top-level name with a project name, usually a client's name, and then with the name of a particular site or high-level project we're building. Themes and plugins written for the same client should share a namespace. If we have two identically-name classes or functions in the same namespace, that's a sign there might be duplicate code we can get rid of.
+  
+When we manage multiple sites for a single client and create a _shared plugin_ for common functionality, it should have its own namespace named "Common" (example: TenUp\Tyrell-Corporation\Common").
```
tlovett1: Seems like we need to be more specific than `Common` as many projects have multiple shared plugins.

johnpbloch: Personally, I don't think common plugins should be treated any different from themes. For example, if `Tyrell_Corporation` has a Twitter plugin named "twitter", the namespace should be `TenUp\Tyrell_Corporation\Twitter` because that is the "code space" proper to that set of code. If this is an internal project, it seems to me we could skip the "client" part and just do "vendor\project" (e.g. `TenUp\ElasticPress`).

johnpbloch:
The sample code in the PR needs some attention before merge. Namespaces follow the same naming rules as classes, so `Tyrell-Corporation` should be `Tyrell_Corporation`, for example.

dana-ross:
And there's also the fact nobody under 30 has seen Blade Runner :(

johnpbloch:
> And there's also the fact nobody under 30 has seen Blade Runner :(

tumblr_llkfapg9cu1qf9jcfo1_500

tlovett1:
@ericmann thoughts on this one?

morganestes:
I'm curious why we don't want to namespace template files. I understand that we shouldn't be defining functions in those files, but if we're using namespaces throughout the codebase, why exclude those particular files? 

:+1: to the note that namespaces should be generally defined _and documented_ for a project ahead of time. That's sage advice for preventing headaches down the road.

johnpbloch:
In PHP, namespaces only affect definitions of constants, functions, classes, interfaces, and traits. None of those things should be defined in a template file (separation of logic/display, etc.). One _could_ make the argument that namespacing the template would reduce the verbosity of the template by placing it in the same namespace as a collection of template functions, but I find that argument unconvincing as it tends to only increase confusion about which function is being used and you can approximate the concision by a well placed `use` statement anyway.

nicoladj77:
I know I'm a little late for this, I missed this discussion earlier. 
One thing to note about `use` statements is that they depend a lot on how you structure and name your classes.
Let's say you have 

```
use TenUp\Events\Page;

// 150 lines below
$page = new Page();
```

in this case the `use` statement just adds confusion, you'd be better off using

```
use TenUp\Events;

// 150 lines below
$page = new Events\Page();
```

One more thing to consider is that using namespaces which mirrors the folder structure means that you can use a PSR compliant autoloader like composer. 
I agree that template tags should not be namespaced.

PR #97, opened 2015-08-15
Avoiding hooks in class constructor

ryanwelcher:
Adding a note about avoiding adding hooks in the constructor of a class and suggesting a separate init method. Not sure if this is too prescriptive but I think it should be noted because it makes unit testing the code much more difficult as it would require WordPress to be loaded or the actions to be mocked.


tlovett1:
Totally agree.

Issue #92, opened 2015-07-22
(here|now)doc syntax


scottlee:
This topic came up internally recently and we thought it'd be worth pitching here as well: **should we consider offering advice on how, when, and if to use (here|now)doc syntax?**
## Take this code as example:

```
$image_html = <<<HTML
<div class="entry-thumbnail">
    <a href="$permalink>
        <img src="$image_src" />
    </a>
</div>
HTML;
return $image_html;
```
### Pro/Con
- Pro: Easy to read. (vs. `?><!-- some html here --><?php` )
- Con: Doesn't allow for truly **late** escaping.
- Con: Increased chance of mistake. One typo could cripple the code below.
- ??

CC: @daveross, @zrothauser

mphillips:
Another option is to use output buffering. The syntax seems cleaner if you need to call functions or or add in other logic.

``` php
<?php
ob_start(); ?>

<div class="entry-thumbnail">
    <a href="<?php echo esc_url( $permalink ) ?>">
        <img src="<?php echo esc_url( $image_src ) ?>" />
    </a>
</div>

<?php
$image_html = ob_get_contents(); 
ob_end_clean(); ?>
```

But, I'm not sure we should standardize putting HTML in a variable because it is less secure and you can't escape late (right before output). Of course, there are edge cases that require putting HTML in a variable, but I don't think we should get into the habit of doing it frequently.

Sorry to butt in. :)

scottlee:
You're totally not butting in, @mphillips. You're spot on, output buffering is very much in the same vein as *doc syntax.

nicoladj77:
I usually use Heredoc and call any function i need before it because it avoids to enter exit/php, I really have hard times reading php inside html code. I think it's more readable than double quotes as you can avoid escaping double quotes. It basically becomes a logic-less templating engine.
I think that heredoc syntax is ok for 5-6 rows, not for bigger chunks of html.

morganestes:
I like heredoc for the ease of reading, but the limitations of no late escaping and no processing inside (can't run a function call inside a heredoc) makes it pretty much unusable for the majority of WordPress use-cases. 

I think the no late escaping is problematic enough on its own to warrant not including heredoc as a recommended practice, but I do see value in including it as an alternative method with reasons to avoid the use of it.

nicoladj77:
[...]

PR #100, opened 2015-08-24
First stab at a note on heredoc/nowdoc syntax ( issue #92 )

dana-ross:
Collaborated with @scottlee on this. Feedback/edits welcome.

### DIFF
--- a/_includes/markdown/PHP.md
+++ b/_includes/markdown/PHP.md
@@ -709,7 +709,7 @@ However, heredoc/nowdoc make it impossible to practice *late escaping*:
 $a = esc_attr( $my_class_name );
 
 // Something naughty could happen to the string after early escaping
-// $a .= 'something naughty';
+$a .= 'something naughty';
 
 $x = <<<HTML
 <div class="test {$a}">test</div>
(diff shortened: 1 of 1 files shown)


tlovett1:
I think this looks great. Not sure how I feel about the `get_template_part` section though. I think we should be using that a bit more judiciously. I'm not a fan of opening a theme to find 30 template partials. Thoughts?

ericmann:
As much as I don't like having to grok lots of template partials, dividing the theme itself up into multiple, purpose-specific partials does make sense (particularly if they're re-usable).

Usually, the argument for a heredoc/nowdoc block is that it's a chunk of HTML (or other non-PHP) code that needs to be reused elsewhere. That's the same reason we break out smaller template parts. Some VIP sites we've worked on have rich template directories with over 100 different template partials (due to the way we've broken the site down into individually-styled, standalone, reusable components).

So I can totally see the use of `get_template_part()` in place of heredoc/nowdoc syntax for large chunks of markup. I think either presenting or linking to an example of how that can be achieved would make the advice even stronger.

dana-ross:
@tlovett1 @ericmann What do you think, should we merge this? Happy to take out the get_template_part() section if that's a sticking point.

tlovett1:
I'm good with this.

PR #113, opened 2015-10-17
Fix typos

rclilly:
When I was reading through this a few typos jumped out at me, so I thought I'd help out by fixing them and letting you know about them.


tlovett1:
Thank you!

PR #118, opened 2015-12-04
Fix to bug "esc_js() description is inaccurate"

raphaelkross:
A better description is given to when using esc_js is correct and what use instead when you want to print a safe string to be used in your JS code (wp_json_encode).


tlovett1:
Awesome.

PR #119, opened 2015-12-07
First draft of Imports & long-running processes

dana-ross:
Just a first draft. Hoping to get feedback from @tott and others. Thorston has a lot of good info from his days with VIP.

He suggested using `wp_suspend_cache_invalidation( true )` to disable cache updates while the process is running, but we would need to invalidate and/or rebuild those caches when the long-running process completed. So before I can recommend that as a best practice, I'd need to find every cache affected by this setting.


tlovett1:
@daveross @tott any update?

sc0ttkclark:
I'd like to throw some tweaks to this as part of another PR or additional notes for covering memory management.

timwright12:
@daveross Is this still active or should we close it out?

dana-ross:
@timwright12 Everything in here is either barely relevant (VIP classic migrations) or covered in the Migrations section. Let's close it out.

Issue #123, opened 2015-12-18
Nonce section cleanup


ivankruchkoff:
Improve nonce documentation to include replay attacks, and an example of nonce tied to an action (as per code example) and a nonce tied to an action tied to a unique ID (as per trashing post example)

nprasath002:
I think this is already documented now?

PR #124, opened 2015-12-18
Nonces: replay attacks and nonces tied to action and unique id

ivankruchkoff:
Address #123

### DIFF
--- a/_includes/markdown/PHP.md
+++ b/_includes/markdown/PHP.md
@@ -578,25 +578,24 @@ There are many escaping situations not covered in this section. Everyone should 
 
 #### Nonces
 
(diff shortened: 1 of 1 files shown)


tlovett1:
It would be nice to get some more clarification on "unique" vs. non-unique nonces.

dana-ross:
> WordPress' implementation of nonces are not strictly numbers...
> 
> The WordPress definition of nonces is "A cryptographic token...

@ivankruchkoff I think these paragraphs are similar enough they should be combined. That would also give a good opportunity to clarify how WP nonces don't necessarily _prevent_ replay attacks, but they _do_ limit the window of time for them. The difference might not be immediately obvious.

tlovett1:
Any update here?

timwright12:
@ivankruchkoff Is this still active or should we close it out?

PR #127, opened 2016-01-04
JS Best Practices Configurations

### LINKED ISSUE #126: Editor Config
Using an EditorConfig helps developers define and maintain consistent coding styles between different editors and IDEs. EditorConfig files are easily readable and they work nicely with version control systems. This is especially useful when there are many of us pushing to one staging branch and we want to reduce as many conflicts as possible.

http://editorconfig.org/#download

I have been using these in many projects already and have only had positive experiences. Today, I just noticed it is not included in our best practices. Have you been using `.editorconfig`? Do you think something like my first paragraph could be a useful entry?

jonathantneal:
[EditorConfig](http://survivejs.com/webpack_react/linting_in_webpack/#editorconfig) allows you to maintain a consistent coding style across different IDEs and editors.

Resolves #126


d2s:
Good that you created two separate Git commits, so that the changes are more understandable.

tlovett1:
I was thinking more along the lines of proposing a standard .editorconfig for projects. I am worried about how markdown might display if I merge. Have you tested locally?

jonathantneal:
I am going to change this pull to only include configuration files. That way, it will not conflict with other merges (as this one now has), and it can be referenced by our own documentation on the subject.

jonathantneal:
@morganestes, I’ve added `.editorconfig`, `.eslintrc`, and `.jscsrc` configurations to this pull.

The `.jscsrc` configuration was forked from WordPress with rules from Google and some rules we follow internally (and [in this repo](https://github.com/10up/Engineering-Best-Practices/blob/gh-pages/js/main.js)).

The `.editorconfig` and `.eslintrc` configurations was forked from PostCSS with rules we follow internally.

I want to give credit where it is due. Several of these `.jscsrc` tweaks were only made possible thanks to @hzoo, a core contributor to jscs-dev whom I give massive kudos to. He may remember my headaches with setting up “proper” function spacing.

hzoo:
@jonathantneal haha I do, thanks! Also you probably already know this but you can also write your own eslint/jscs rules/plugins for styles specific your codebase/etc.

tlovett1:
Awesome!

PR #128, opened 2016-01-07
Feature/transient timeout autoload

ryanwelcher:
Adding a note about transients without an expiry being auto-loaded by default.


tlovett1:
Awesome.

PR #129, opened 2016-01-08
Adding JSHint configuration details

allenmoore:
This includes 10up's use of JSHint for code quality, error detection, and conformity to the WordPress JavaScript best practices. This also includes a sample code snippet for JSHint.


tlovett1:
Looks good @allenmoore.

My first thought is that the section should be title something other than `Code Quality` since most of the Best Practices are about that. Would love to see some other people weigh in.

jonathantneal:
We can be specific about the activity:

> ### Linting
> 
> Linting is the process of analyzing code for potential errors. We use...

Or focus on the purpose.

> ### Error Prevention
> 
> We use...

Or cover the big picture.

> ### Code Conformance
> 
> Something about JSHint for error prevention and method conformance...
> 
> Something about JSCS and EditorConfig for code style conformance...

allenmoore:
After talking with @jonathantneal about ESLint versus JSHint, I started testing ESLint. The rule options that ESLint provides is a step ahead of JSHint. In some cases, I had to add a couple of inline ignores. This is very similar to PHPCS. In my opinion, this isn't a blocker for using ESLint. In fact, it would be a good way of gaining a better understanding of how JavaScript scope works and the potential issues that can take place from the code we write. I'd love to hear @jonathantneal's or @daveross's thoughts.

dana-ross:
@allenmoore I don't have a problem with inline ignores as long as they're used intelligently. You have to understand the rules before you can break them.

The engineers I've worked with at 10up have been able to use PHPCS ignores responsibly. They should be fine with JavaScript ones, too.

tlovett1:
I'm fine with just sticking with JSHint for now. Can we change the section title?

tlovett1:
Now I'm thinking we should switch to ESLint.

allenmoore:
@tlovett1 Agreed! I'll make the changes to this pull request replacing JSHint with ESLint. I've created a ESLint config file that we could potentially include as a reference. It can be seen here -- https://github.com/allenmoore/eslint-config.

Gemfile.lock
```diff
@@ -14,34 +13,31 @@ GEM
       execjs
     coffee-script-source (1.10.0)
     colorator (0.1)
-    ethon (0.8.1)
+    ethon (0.9.0)
```
ericmann: Would rather comment on the whole file, but it seems like it would be a good idea to keep Gemfile changes out of feature-specific changes. No argument against upgrading dependencies, but that's not related to the rest of the PR at all.

allenmoore:
@ericmann Great catch! I reverted the commit that included the changes in `Gemfile.lock` and then pushed a new commit with the appropriate changes to page nav weights.

dana-ross:
This PR is very stale at this point. I almost hesitate to close it because I think this is the moment where we made the jump from JSHint to ESLint. It's a little time capsule of 10up history.

The new project scaffolds we're rolling out are built with ESLint support, and I think that's being used universally on 10up projects. I'm not sure there's value in us publishing an `.eslintrc` beyond that (@tlovett1 & @timwright12 , thoughts?).

Closing for now.

timwright12:
I'm good with just using the eslintrc

PR #131, opened 2016-01-11
Update Commit message best practices

### LINKED ISSUE #109: Commit message best practices
I just published [best practices for commit messages in WordPress core](https://make.wordpress.org/core/handbook/best-practices/commit-messages/), much of which is general commit message advice. This is mostly a placeholder issue so we add this to our own best practices at 10up, with some blog posts I plan on writing to go along with it.

jonathantneal:
This gives everyone advice for writing commit messages. It borrows heavily from and resolves #109.

### DIFF
--- a/_includes/markdown/Version-Control.md
+++ b/_includes/markdown/Version-Control.md
@@ -12,7 +12,42 @@ At 10up we consider standardizing a workflow to be a very important part of the 
 
 #### Commits
 
-Commits should be small and independent items of work. Distinct items of work are essential in keeping features separate, which will aid in reversing or rolling back commits if necessary or pushing specific features forward.
+Commits should be small and independent items of work, containing changes limited to a distinct idea. Distinct commits are essential in keeping features separate, pushing specific features forward, or reversing or rolling back code if necessary.
+
+##### Commit Messages
+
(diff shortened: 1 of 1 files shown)


helen:
This is a good start. I think it would be good to include more about how one goes about deciding how much detail to include, e.g. a typical, hopefully-atomic commit while working on something vs. a merge.

tlovett1:
I'd also like to see expectations set on what is required. I think we require the short description but not the long one.

jonathantneal:
I’ve re-written this section to be simpler to read and more clear about our expectations, including examples of quality commit messages alongside our advice.

dana-ross:
Thanks @jonathantneal . Love the examples. I know I'm definitely guilty of blowing past the 70 character limit & spilling details into the commit message.

On that note, maybe add a warning against using the `git -m` command line option?

jonathantneal:
@daveross, I think `-m` has it’s place, and my examples show how and when to use `git commit` with or without the `-m` option. Thanks for the kind review, btw. @tlovett1, is this yet merge-worthy?

tlovett1:
Looks solid to me.

PR #134, opened 2016-01-30
Add Variable VVV for site creation.

bswatson:
I find this to be the easiest methods to create new sites in VVV.

The only concern I have with adding this as a recommendation is the default method for hooking up git repositories is by targeting wp-content, which doesn't follow our standard practice.

### DIFF
--- a/_includes/markdown/Tools.md
+++ b/_includes/markdown/Tools.md
@@ -8,9 +8,11 @@ At 10up, we use [Vagrant](https://www.vagrantup.com/) to build and interact with
 
 [VIP Quickstart](https://github.com/Automattic/vip-quickstart) - A Vagrant setup meant to closely match the WordPress.com VIP environment. If you are working on a VIP project, it is good to have this installed to effectively test your website locally. We only recommend this Vagrant setup for VIP projects.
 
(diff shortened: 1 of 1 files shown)


jonathantneal:
Variable VV is a great tool! I use it all the time.

Why is the default method to hook up git repositories to wp-content anyway. I do see this in many of our projects. When you say it doesn’t follow our standard practice, are you referring to this:

> If we are building a theme, we version control only the theme and deploy to the themes directory. — [Version Control, Structure](https://10up.github.io/Engineering-Best-Practices/version-control/#structure)

bswatson:
@jonathantneal - The's the line I'm referencing, and from experience the preferred method of handling a new build. We do have many projects that use the wp-content folder as the root, but there are project specific reasons in those cases.

I'd like to point out that, without any command line argument, Variable VVV doesn't hook up to a git repo, so you'd need to specify the argument to pull down a repo as part of the provisioning. However, the only option for git integration is to have it dropped on wp-content.

jonbellah:
Agree with adding Variable VVV as a recommendation. Disagree with changing 'task runner' to 'taskrunner' in the PR.

bswatson:
@JonBellah - Removed that change. Not sure how it made it's way in unless my IDE decided to autocorrect spelling in the file.

PR #137, opened 2016-03-18
Switch to Kramdown markdown engine and Rouge syntax highlighter

zrothauser:
Updates the site to use the new default Github Pages markdown engine and highlighter, and adjusts the indentation of some of the code samples to work with the new markdown engine.


tlovett1:
Thanks Zack!

PR #143, opened 2016-07-03
Feature/add editor config section

dsawardekar:
@tlovett1 @morganestes I took a stab at this. I've added a section with details about where to download the Editor config plugin and the default `.editorconfig` adapted from Core.

Let me know if you'd like to see any changes.


ericmann:
There's some content in this .editorconfig that doesn't exist in the wp-make template, and vice versa. I think it would be a good idea to set the standard _here_ then update Make to match. I've opened an issue over there as well ... https://github.com/10up/generator-wp-make/issues/75

dana-ross:
@tlovett1 How do you feel about merging this?

PR #144, opened 2016-07-03
Adds section on decoupling plugin and theme code

dsawardekar:
@tlovett1 This PR adds recommendations on using `add_theme_support` to help decouple plugin and theme functionality. It includes a simple example where custom JS is added to a page, only if the theme registers support for it.


tlovett1:
Awesome!

PR #145, opened 2016-08-02
Added Default HTML standard for accessible tabs

timwright12:
I updated the tab markup with more fleshed out aria and accessibility features. Base reference: http://simplyaccessible.com/article/danger-aria-tabs/


tlovett1:
Sweet!

PR #156, opened 2016-10-06
Fix menu nav uneven alignment

RachelRVasquez:
Looks like the issue we have on the live io site was for some reason, it was no longer understanding the sort by weight. Changed to sort by name, which seems like what Github was doing instead, except changed the file names so we can control the order.


tlovett1:
This is producing an error: 

The page build failed with the following error:

A file was included in `css.md` that is a symlink or does not exist in your `_includes` directory. For more information, see https://help.github.com/articles/page-build-failed-file-is-a-symlink.

PR #159, opened 2016-10-13
Add a hyphen to "WP-CLI"

stevegrunwell:
Per http://wp-cli.org, WP-CLI is written as a hyphenated word, not "WP CLI".


tlovett1:
Going to close this one for now since we are re-writing a lot of this section. Sorry!

PR #161, opened 2016-10-13
Fix anchor link title visibility

RachelRVasquez:
Made fixes so the headings will show when clicking anchor links from the header subnav. Also added in a subnav for Migrations in the process.


tlovett1:
Need conflicts resolved.

dana-ross:
This is a stale PR with conflicts. But it's a small change with plenty of value. I'd love to see someone get it over the finish line, probably as a fresh PR based on the current codebase.

What Rachel did was add a class for the headings with `padding-top: 60px; margin-top: -60px;` and then add that class to all the heading tags. This has no visible effect, except when you reference a heading ID in the URL (i.e. https://10up.github.io/Engineering-Best-Practices/php/#caching) the header of that section isn't scrolled up off the top of the screen.

Great one for someone looking to make their first contribution to the Engineering Best Practices.

dana-ross:
Re-created as part of #226

Issue #136, opened 2016-03-17
Versioning css and js files to bust CDN and browser cache?


TheLastCicada:
I skimmed through and I didn't find anything about how to version js and css code in WordPress to bust CDN and browser caches when deploying updates.  Something like adding ?ver=0.1.5 to the end of the enqueue URL so that the CDN picks it up as a new asset.  I see this as confusing for new engineers all the time and it'd be great to have a best practice to point to.

tlovett1:
Agree this should be required. WP actually supports a parameter for versioning assets.

jonbellah:
Agree.

Generally, we use the `{THEME}_VERSION` constant, found by default in wp-make, and attach it to all of the assets we're loading through `wp_enqueue_script()` and `wp_enqueue_style()`.

If no one has an issue with that approach, then I write up a quick PR.

tlovett1:
Sounds right to me.

stevegrunwell:
@TheLastCicada have we made progress getting AAD available publicly? It would be great to be able to skip the version constants altogether and instead dynamically bust caches for assets that aren't given explicit versions.

PR #163, opened 2016-10-18
Add the Asset Versioning section

jonbellah:
Covers a basic process for keeping assets properly versioned. Addresses #136

### DIFF
--- a/_includes/markdown/PHP.md
+++ b/_includes/markdown/PHP.md
@@ -521,6 +521,20 @@ if ( current_theme_supports( 'custom-js-feature' ) ) {
 }
 ```
 
+#### Asset Versioning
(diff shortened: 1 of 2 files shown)


_includes/markdown/PHP.md
```diff
@@ -521,6 +521,20 @@ if ( current_theme_supports( 'custom-js-feature' ) ) {
...
+It is recommended that engineers use a constant to define their theme or plugin version, then reference that constant when using registering scripts or styles. For example:
+
+```php
+<?php
+define( 'THEME_VERSION', '0.1.0' );
+
+wp_register_script( 'custom-script', get_template_directory_uri() . '/js/asset.js', array(), THEME_VERSION );
+```
+
+Remember to increment the version in the defined constant prior to deployment.
```
stevegrunwell, line 536: It's probably worth adding a disclaimer that assets that already have real versions (e.g. version 0.5.0 of js/vendor/my-cool-library.js) should use the version number from the library, not the theme asset version. This avoids the cache from needlessly being flushed when underlying vendor script hasn't actually changed.

ericmann: 👍

jonbellah: Good thinkin'... I'll add that.

tlovett1:
Nice!

PR #164, opened 2016-10-18
Expand the Social section into an External Integrations section

dkotter:
Adding more details around not triggering external integrations when running migrations. The previous section was called Social but have changed that to be External Integrations, as there are other things that could be triggered besides social items.

Also moved this section up a bit to be more prominent, and added details about using the `WP_IMPORTING` constant in that section (as well as left it later in the doc).


_includes/markdown/Migrations.md
```diff
@@ -36,6 +36,18 @@ The following are some general things to keep in mind when doing these migration
 
 	Sometimes media can be left as-is, meaning, for instance, any image that's referenced somewhere can stay as-is and doesn't need to be moved over. But often we'll want all media brought over to the new site. In this case, if the amount of total media items is small, this can be done using core WordPress functions. If the amount of media is great, this can significantly slow down the migration process and sometimes crash it all together. In this scenario, looking into options to offload this processing is usually the way to go (something like using [Gearman](http://gearman.org/) to process multiple media items at once).
 
+* *External integrations*
```
ericmann, line 39: Potentially cross-link to the existing ["Third Party Integrations" section](https://10up.github.io/Engineering-Best-Practices/structure/#integrations) as they serve the same purpose?

tlovett1: @dkotter can you add a link? Thanks @ericmann !!

PR #172, opened 2017-01-16
Adding in Gulp, Browserify, and Webpack information.

fuhton:
Expanding the Task Runner section to include Gulp/Browserify/Webpack.

### DIFF
--- a/_includes/markdown/Tools.md
+++ b/_includes/markdown/Tools.md
@@ -12,7 +12,13 @@ At 10up, we use [Vagrant](https://www.vagrantup.com/) to build and interact with
 
 <h3 id="task-runners">Task Runners {% include Util/top %}</h3>
 
-[Grunt](http://gruntjs.com/) - Grunt is a task runner built on Node that lets you automate tasks like Sass preprocessing and JS minification. Grunt is the only task runner we currently use on company and client projects.
+[Grunt](http://gruntjs.com/) - Grunt is a task runner built on Node that lets you automate tasks like Sass preprocessing and JS minification. Grunt is our default task runner and has a great community of plugins and solutions we use for on company and client projects.
+
+[Gulp](http://gulpjs.com/) - Gulp is also a task runner build on Node that offers a similar suite of plugins and solutions to Grunt. The biggest difference is Gulp allows you direct access to the [stream](https://nodejs.org/api/stream.html) of information from your source files and allows you to modify this data directly.
+
+[Browserify](http://browserify.org/) - Browserify is a bundler that allows us to include Node modules in our JavaScript files. This tool speeds up development and can reduce the amount of WordPress enqueue's we need for different vendor files or other JavaScript requirements.
+
+[Webpack](https://webpack.github.io/) - Webpack is a bundler/task manager - the next generation of JavaScript/CSS development tools. At 10up we are actively building new projects with this library to expand our internal knowledge and find the best use cases. Webpack allows us to combine all of good parts of the above mentioned tools into a single file/command.
 
 <h3 id="package-managers">Package/Dependency Managers</h3>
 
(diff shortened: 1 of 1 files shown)


ericmann (approved):

tlovett1:
Not sure how I feel about adding Browserify. I think Gulp makes sense as it's really replaced Grunt. The main reason we've stuck with Grunt is for consistency and the fact that WP still uses it. However, Browserify seems like it's largely being overtaken by Webpack. Thoughts?

fuhton:
@tlovett1 I've been wrestling with this as well, but wanted to cover my bases. A year ago, I think Browserify might have made sense, but now (in client projects and personal projects) I'm recommending Webpack out of the box. I'm okay dropping that section.

_includes/markdown/Tools.md
```diff
@@ -12,7 +12,11 @@ At 10up, we use [Vagrant](https://www.vagrantup.com/) to build and interact with
...
 
-[Grunt](http://gruntjs.com/) - Grunt is a task runner built on Node that lets you automate tasks like Sass preprocessing and JS minification. Grunt is the only task runner we currently use on company and client projects.
+[Grunt](http://gruntjs.com/) - Grunt is a task runner built on Node that lets you automate tasks like Sass preprocessing and JS minification. Grunt is our default task runner and has a great community of plugins and solutions we use for on company and client projects.
+
+[Gulp](http://gulpjs.com/) - Gulp is also a task runner build on Node that offers a similar suite of plugins and solutions to Grunt. The biggest difference is Gulp allows you direct access to the [stream](https://nodejs.org/api/stream.html) of information from your source files and allows you to modify this data directly.
+
+[Webpack](https://webpack.github.io/) - Webpack is a bundler/task manager - the next generation of JavaScript/CSS development tools. At 10up we are actively building new projects with this library to expand our internal knowledge and find the best use cases. Webpack allows us to combine all of good parts of the above mentioned tools into a single file/command.
```
lkwdwrd: I don't know if webpack is specifically a task runner in the traditional sense, is it? It's a bundler and you can run different file types through it, but you can't really define an arbitrary task to run, right?

jonbellah: In cases where you use Webpack as a bundler, it generally makes sense to just use plain `npm scripts` to manage 'tasks'. Seems to be the flavor-du-jour.

fuhton: Good catch. Was in the "task manager" frame of mind.

_includes/markdown/Tools.md
```diff
@@ -12,7 +12,11 @@ At 10up, we use [Vagrant](https://www.vagrantup.com/) to build and interact with
...
 
-[Grunt](http://gruntjs.com/) - Grunt is a task runner built on Node that lets you automate tasks like Sass preprocessing and JS minification. Grunt is the only task runner we currently use on company and client projects.
+[Grunt](http://gruntjs.com/) - Grunt is a task runner built on Node that lets you automate tasks like Sass preprocessing and JS minification. Grunt is our default task runner and has a great community of plugins and solutions we use for on company and client projects.
+
+[Gulp](http://gulpjs.com/) - Gulp is also a task runner build on Node that offers a similar suite of plugins and solutions to Grunt. The biggest difference is Gulp allows you direct access to the [stream](https://nodejs.org/api/stream.html) of information from your source files and allows you to modify this data directly.
+
+[Webpack](https://webpack.github.io/) - Webpack is a bundler - the next generation of JavaScript/CSS development tools. At 10up we are actively building new projects with this library to expand our internal knowledge and find the best use cases. Webpack allows us to combine all of good parts of the above mentioned tools into a single file/command.
```
tlovett1: Not sure I understand what this means: "actively building new projects with this library to expand our internal knowledge and find the best use cases"

Issue #162, opened 2016-10-13
Package/Dependency Management with Yarn


jonbellah:
A couple of days ago, Facebook announced a new package manager called [Yarn](https://code.facebook.com/posts/1840075619545360/yarn-a-new-package-manager-for-javascript/). Yarn allows engineers to continue utilizing the npm registry, but has some additional performance and security considerations built in.

Yarn also has the potential to be a single solution for javascript and front-end package/dependency management, [bridging the gap between Bower and npm](https://bower.io/blog/2016/using-bower-with-yarn/).

I know it's still new and fairly unproven, but I figured I would at least start the conversation.

joshlevinson:
I've been using yarn since shortly after its release. Faster installs, no issues for me locally.
I've modified the 10up Redux DeployBot repo to use yarn; it's working great there!
I'll likely continue to use yarn on new DB setups. 👍 from me!

tlovett1:
Yarn looks really promising. I realize it uses npm behind the scenes but it seems too early to all out replace npm.

jonbellah:
Bumping this thread, now that Bower has officially been deprecated: https://github.com/bower/bower/issues/2298

Will put together a PR soon to remove Bower from our toolset, with a separate PR to include Yarn.

ericmann:
As a Bower contributor, this makes me sad :'( But @JonBellah, that's definitely the way to go.

timwright12:
Closing this out to file a new issue around updating the docs to reflect NPM for dependency management

PR #173, opened 2017-01-23
Change bower out for NPM and include a broader JavaScript use case

fuhton:
Going with the trend of modern web development, I think it makes sense to drop our section for Bower and replace it with a section about NPM. I'd also like to mention that dependency managers are great for polyfills and using modern objects - like Promises. This shows that 10up supports and encourages finding modern solutions that support all required browsers.


ericmann:
There was a time when we were trying to make an intentional separation between the different package managers being used:

- Composer was for backend (PHP) dependencies
- Bower was for frontend (JS and CSS/SASS) dependencies
- NPM was for development (build and linting) dependencies

As a Bower maintainer, I'm a bit biased when it comes to removing it from a best practices doc, but I also understand the argument for simplifying things. That being said, if changes are being made to the build tools documentation they should stay consistent in approach.

What I mean by that:

1. There needs to be a very clear delineation between build tools and runtime tools. Some build systems (like DeployBot) will run `npm install` so it can also run linting/minification tools to build a project before deploying it to a server. DeployBot specifically is often configured to _not_ deploy the `.node_modules` directory as it contains a whole slew of content not meant to be run in production. If NPM is going to be used both for build _and_ runtime, there needs to be a way to separate these dependencies both in terms of coding (`dev_dependencies` vs `dependencies`) _and_ during automation (i.e. a well-documented install/build/clean/install/deploy workflow) to avoid deploying the wrong code.
2. There's already discussion on #162 about using Yarn as an alternative to NPM. I, for one, would think it'd be better to keep vanilla NPM around for build-time tools and use something like Yarn (which uses NPM transparently) for frontend runtime dependencies (thus replacing Bower) as it solves the issue above.

_includes/markdown/Tools.md
```diff
@@ -16,7 +16,7 @@ At 10up, we use [Vagrant](https://www.vagrantup.com/) to build and interact with
 
 <h3 id="package-managers">Package/Dependency Managers</h3>
 
-[Bower](http://bower.io/) - A good tool to manage front end packages. Usually everything we need is bundled with WordPress. Sometimes we need something like "Chosen.js" that isn’t included. Bower is a good way to manage external libraries like that but is not necessary on most projects.
+[NPM](https://www.npmjs.com/) - NPM is a great tool to manage front end dependencies. Usually everything we need is bundled with WordPress, but sometimes we need something like "Chosen.js" or [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) support for all browsers. NPM is a good way to manage external libraries or polyfills like these but are not necessary on most projects.
```
tlovett1: Probably worth mentioning npm is good for more than just front-end deps

tlovett1:
Would like to hear more opinions. This seems right to me.

fuhton:
@ericmann I still think the right course of action is to drop the Bower section in favor of NPM. You do make some great points about noting the difference between dependencies and devDependencies - a point a lot of people miss their first few (or many) times with NPM. I've pushed a change to this PR that talks to this a little more and offers the NPM docs as a reference before diving into NPM. I don't see a lot of value in documenting the same thing in 2 places, but have taken the leap to broadly represent 10up's use of the tool and pushed people to the NPM docs for education.

Regarding Yarn vs NPM - to me it's a mostly moot point. Yarn can be a wholesale replacement for NPM - from dependency management to build-time tools, but first it still relies on NPM being understood and used properly. Saying that we want to use one for build tools and one for dependency management is confusing to me. The point should be to simplify our build/dependency tools, not create more rules for when to use one vs the other. Representing that in our best practices is the first step in education. For our deployments, we can setup the environment to have Yarn installed much like we have NPM installed on our deployment environments and instantly you can swap the two out, further making NPM education a priority.

jonathantneal:
You may want to clarify when one should use dependencies or devDependencies. Otherwise, it might be confusing to someone newer to package management.

For instance, a developer might have a theme or a plugin using a Promise polyfill, and they might ask themselves, “Is this polyfill package a `dependency` or a `devDependency`?”. They might figure, “Because the polyfill gets bundled with my proejct during the build process, that makes it a `devDependency`. After all, the build tool required to bundle this is a `devDependency`, right?”.

On the other hand, a developer might realize that there’s a chance this theme or plugin gets used as a dependency by some *other* theme or plugin, in which case that polyfill needs to come along for the ride. In this later example, the developer only realized it’s a dependency because they understand a concept I recommend you clarify: *dependencies get installed alongside the project wherever it gets used. devDependencies only get installed when you’re working on the project itself.*

You might be able to word that better than me, and please correct me if you see things differently.

ericmann:
Let me step back a bit.

Whether or not the docs drop Bower for NPM isn't really an issue. Bower is (sadly) in decline and forward-looking development should both set and use community standards. That said, there are serious ramifications to using the same tool for both dev and production dependencies.

`npm install` by default installs _both_ sets to the `.node_modules` directory. This is fine for local development (or even for staging) and definitely what's appropriate for a build server like DeployBot. However, using NPM for both dev and prod dependencies means a deployment process becomes:

1. Clone the repo
1. `npm install`
1. Build the project
1. `rm -rf .node_modules`
1. `npm install --production`
1. Copy the built project to the server

This is a greater number of steps than the current, separated process would involve. While "number of steps" is an arbitrary metric, keep in mind that doing things _the right way_ should be the easiest because developers move fast while setting up deployments and will want to follow the easiest process. If steps 4 and 5 are skipped, this will result in deploying several MBs of unnecessary code to the server.

For a specific example, one project we built early last year that deployed through DeployBot accidentally included all of the build deps from `.node_modules`. Every deployment with that configuration took 45-50minutes. Once we purged the build dependencies (i.e. ignored the `.node_modules` directory) our deployment time dropped to 2-3 minutes. **This point is critical.**

> The point should be to simplify our build/dependency tools, not create more rules for when to use one vs the other.

I agree with that aim, but using the same tool for both build dependencies _and_ front-end dependencies runs the risk of complicating things in practice, even if it feels simpler in theory.

fuhton:
@jonathantneal 
Good catch. I'll add that in this weekend!

@ericmann 
I totally hear your concerns on deployments and sorry I didn't address that more clearly in my last response! I wanted to be sure the dependency management section was thoroughly clear. Deployments are one of the biggests parts of a development workflow and keeping them quick (and secure) is one of the top priorities. I think you and I are generally in agreement on deployments, so I think our suggestions would benefit the Best Practices overall. Maybe the right solution is to include a new section specifically about Deployments - what do you think? Modify this current PR to include a deployment section and have most of our discussion represented in that new area?

Points to consider for this new section
* Deployment Tools or "Ways to Deploy Code" (atomic {Deploybot}, SFTP uploads{Beanstalk}, WebHooks {Github}, etc..)
* Best practices for prod/development dependencies
* Maintaining security for deployments
* Enqueueing and concatanation - Ending up with only what you need on production. (DLL's, bundled-compiled files, and files usage)

My question to you is - have we covered all uses of this tool (NPM for dependency management)? If we suggest this for modern web development, does a section on dependency management and deployment solutions cover the use cases of NPM?

I'm gonna work on adding in a deployment section to the PR this weekend/early next week.

_includes/markdown/Tools.md
```diff
@@ -26,6 +26,21 @@ At 10up, we use [Vagrant](https://www.vagrantup.com/) to build and interact with
 
 [SVN](https://subversion.apache.org/) - We use SVN, but only in the context of WordPress.com VIP. Again, we encourage people to use the command line as we do not support GUI's internally.
 
+<h3 id="deployments">Deployments {% include Util/top %}</h3>
+The best deployment is quick and requires no site down time. At 10up we’ve spent a lot of our efforts to refine this process and make sure our deployments are secure and speedy. Our 3 primary deployment processes are outlined below
+
+* Atomic
```
lkwdwrd, line 32: Each of these should have a short description of what it is

_includes/markdown/Tools.md
```diff
@@ -26,6 +26,21 @@ At 10up, we use [Vagrant](https://www.vagrantup.com/) to build and interact with
...
+* Webhooks
+* SFTP uploads
+
+All of these deployment tools are utilizing a remote VCS repository - Beanstalk, Github, Gitlab, etc... 10up's recommended deployment cycle is to trigger deployments on updates to a remote repository. This trigger will start the deployment commands. It's helpful to have an environment already setup to run your build commands. If this is not the case, you will need to install your development dependencies to run your commands at this time, which will slow down your deployment. After you have the development dependencies installed, run your build commands. Once the correct files have been compiled, remove any dependencies. Build dependencies are no longer needed and dependencies are now represented in your compiled files.
```
lkwdwrd, line 36: The first part here is a bit hard to follow. I believe we're talking about things that are the same for all of the techniques. We could probably make that it's own paragraph to make that clear. Probably a statement that says more directly deployment methods, no matter what's used, should consider do these things...

The part about dev dependencies should be it's own paragraph as well and stress that devDependencies should not go to production. In fact `node_modules` should not go to production, imho. If we use a front end dependency through NPM and it should not be bundled, the build tool should pull it out of the `node_modules` folder and it should be included from the copied location.

_includes/markdown/Tools.md
```diff
@@ -26,6 +26,21 @@ At 10up, we use [Vagrant](https://www.vagrantup.com/) to build and interact with
...
+
+While 10up doesn't recommend one solution or company over another, we do recommend using the best tools and solutions for the job, and that can vary between projects.
```
lkwdwrd, line 38: Would it be worth including the tools we _do_ use?

_includes/markdown/Tools.md
```diff
@@ -26,6 +26,21 @@ At 10up, we use [Vagrant](https://www.vagrantup.com/) to build and interact with
...
+
+For security of your deployment, it's important to follow the above recommendations for dependency management during deployments. When using an SFTP solution, don't use the root user - create a custom ftp user with clearly defined access. You shouldn't share this access with other users. Finally, only give deployment access to those who need it. Not every developer on a team will need access to deployments or the ability to modify those deployments, limiting access to those who absolutely need it.
+
+When deploying code and using the WordPress enqueue system, it's imperative to only have relevant code on the page. Sometimes there is a special page or feature that uses an extra library and instead of requiring this on every page, rely on WordPress to enqueue the file for you. Instead of compiling all your dependencies needed by any feature on your site into one file, compile a few core libraries or tools for use on every page and only enqueue what you need on any given page. This will reduce page load, simplify architecture by making it clear where dependencies get used, and avoid conflict with other plugins or themes on your site.
```
lkwdwrd, line 42: I like this section, but it applies even outside of the WP enqueue system.

lkwdwrd:
I'm gonna contribute a couple of thoughts here, though I'm further from the day to day dependency management and deployment workflows now. I would love @JonBellah to do some thinking on this one as well since he is a lot closer to our front end dependency workflow.

 - In my opinion, we should not deploy `node_modules` folders at all. If we need a dependency it should be copied out of the folder into a well known location within the project and used from there. Even when a compiled file is available in and NPM dependency, there is usually a lot of other files that come along with it. I'd rather not deploy the modules themselves when I'm only interested in a CSS file or JS file within the module. By never deploying `node_modules` it does perhaps make it more annoying for engineers, but hopefully ensure we don't set up a deploy, install megabytes worth of devDependencies, and the deploy them to production because we forgot to `rm -rf node_modules && npm install --production` in the deploy script.
	- This can be a bundled file but doesn't have to be
	- This can be a file copied into some kind of vendors directory
 - A deployments section is a good idea. We've got several methods we use to deploy right now, but in talking with @TheLastCicada we're making some attempt to standardize on DeployBot. I don't know if we should necessarily call that out here, but it's worth some thought.

tlovett1:
I'm on the fence about deployments. I totally see the value in some of the information. However, it seems like the section should be a part of a bigger dev-ops best practices. Would love to hear @TheLastCicada's thoughts.

I think we are mostly in agreement on this thread:
* Bower is dying.
* Use npm for dev/prod dependences.
* Yarn is promising but, since it works the same as npm, not worth diving into when we are only talking about which dependency manager to use.

I think the ideal workflow is:

1. Code pushed to remote
1. npm install
1. Build project
1. [npm prune --production](https://docs.npmjs.com/cli/prune)

_I don't want us to get off topic here. Let's remove Bower and clarify how we use npm with dev/prod dependencies. Deployment section should be moved to a new issue._

PR #175, opened 2017-01-27
Added PHP section on sessions

dana-ross:
In light of recent discussions about session storage, @TheLastCicada & I collaborated on an addition to the PHP page.


_includes/markdown/PHP.md
```diff
@@ -833,4 +833,14 @@ $my_class_name .= 'something naughty';
...
\ No newline at end of file
+Even better, [use WordPress's ```get_template_part()``` function as a basic template engine](http://codex.wordpress.org/Function_Reference/get_template_part#Passing_Variables_to_Template). Make your template file consist mostly of HTML, with ```<?php ?>``` tags just where you need to escape and output. The resulting file will be as readable as a heredoc/nowdoc block, but can still perform late escaping within the template itself.
+
+### Avoid Sessions
+
+Sessions add extra complexity to web sites and extra burden on hosting setups. Avoid using sessions to store individual users' preferences or other data. WordPress VIP [explicitly forbids sessions in their code reviews](https://vip.wordpress.com/documentation/vip/code-review-what-we-look-for/#session_start-and-other-session-related-functions).
```
ericmann, line 840: Perhaps an explicit note that the use of server-side sessions required bypassing any sort of front-end cache?

PR #184, opened 2017-03-06
Various prettifications to unify document style

noplanman:
This PR is basically a bunch of little fixes to make the general style of the pages nicer.

Also a fix here and there that broke the way the document is rendered.


tlovett1:
Closing this one. Prettification is nice, but not super necessary. Thanks!

PR #115, opened 2015-10-28
Promote internationalization by default

DrewAPicture:
I've stubbed out an initial pass at adding default internationalization recommendations as suggested in #114. Work in progress.

PR #190, opened 2017-04-11
Finishing "Promote internationalization by default"

dana-ross:
Continues the work began in #115


tlovett1:
Let me know when this is ready for review. Thanks!

dana-ross:
I think it's ready for review. The only change I'm considering at this point is to expand on the "internationalization by default" in the pull request title. I don't feel the intro gives strong enough instruction to do it.

_includes/markdown/PHP.md
```diff
@@ -616,6 +616,51 @@ if ( ! empty( $_POST['_wpnonce'] ) && wp_verify_nonce( $_POST['_wpnonce'], 'my_a
...
+
+Each project should leverage a unique text domain for its strings. Text domains should be lowercase, alphanumeric, and use hyphens to separate multiple words: `tenup-project-name`.
+
+Text domains should never be stored in a variable or named constant when used with core localization functions, as this practice can often produce unexpected results. Most tools used to create translation rely on [GNU gettext](https://www.gnu.org/software/gettext/) scanning source code for translation functions. PHP code won't be interpreted, only scanned like it was a block of plain text, and translation tools won't be able to assign the text domain correctly if it's not there in plain text.
+
+```php
+// Always this
+$string = __('Hello World', 'plugin-domain');
+// Never this
+$string = __('Hello World', $plugin_domain);
```
tlovett1: I think what you are meaning to say is:

```php
\\ Always this
__('Hello World', 'plugin-domain');

\\ Never this
$string = 'Hello World';
__( $string, 'plugin-domain' );
```

Also please make sure to use proper spacing: `__( 'Hello World', 'plugin-domain' );`

ericmann: I think both. Not having the string literal on either side can break POEdit parsing (it can't find the string to translate or it can't find the textdomain to use in translation).

dana-ross: Eric's correct. `gettext` just isn't smart enough to parse PHP variables anywhere, it's just pattern-matching. Mark Jaquith put out a post that opened my eyes to this when I started out writing plugins: https://markjaquith.wordpress.com/2011/10/06/translating-wordpress-plugins-and-themes-dont-get-clever/

I wonder if I should link to that post, too…

dana-ross: Ok, I pushed up a change that addresses variables inside translated strings (with examples) and tweaked the wording for the same under "text domains" so it's not just a repeat of what's above.

_includes/markdown/PHP.md
```diff
@@ -726,6 +726,65 @@ if ( ! empty( $_POST['_wpnonce'] ) && wp_verify_nonce( $_POST['_wpnonce'], 'my_a
...
+```
+
+Localizing a project differs from the core approach in two distinct ways:
+* A unique text domain should be used with all localization functions
+* Internationalized output should always be escaped
+##### Text Domains
+
+Each project should leverage a unique text domain for its strings. Text domains should be lowercase, alphanumeric, and use hyphens to separate multiple words: `tenup-project-name`.
+
+Like the translated strings they accompany, text domains should never be stored in a variable or named constant when used with core localization functions, as this practice can often produce unexpected results. Most tools used to create translation rely on [GNU gettext](https://www.gnu.org/software/gettext/) scanning source code for translation functions. PHP code won't be interpreted, only scanned like it was a block of plain text, and translation tools won't be able to assign the text domain correctly if it's not there in plain text.
```
tlovett1: " Most tools used to create translation rely on [GNU gettext](https://www.gnu.org/software/gettext/) scanning source code for translation functions." that exact sentence appears twice.

tlovett1: @daveross see this?

dana-ross:
@tlovett1 I finally got around to fixing that duplicated sentence. Take a look when you get a chance & merge it if you think it's ready.

tlovett1:
Looks good.

PR #192, opened 2017-05-16
Add switch_to_blog limitation

kierahowe:
I thought it would be good to mention that switch_to_blog is generally a bad idea on the front-end and is a performance issue if used.

### DIFF
--- a/_includes/markdown/PHP.md
+++ b/_includes/markdown/PHP.md
@@ -148,6 +148,9 @@ if ( isset( $array['bar'] ) ) {
 
 In case you don't have control over the array creation process and are forced to use `in_array()`, to improve the performance slightly, you should always set the third parameter to `true` to force use of strict comparison.
 
+#### Never use switch_to_blog on the front-end
+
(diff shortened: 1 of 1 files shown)


nicholasio:
Last time and checked on this, the biggest problem was more memory consumption that the query itself, I also know that there were major improvements to this function after the initial concerns regarding its performance. I agree that we should generally discourage the use of this function but there are times when it becomes really handy and useful (our Distributor plugin is an example). In my opinion we should, instead of saying "never use this function" say that it's important to have a cache strategy in place and use this function wisely.  For examples I've seen code that does several switch/restore before each template tag function when it could just do one switch/restore and have all functions be called inside that block. We should also recommend pursing other solutions first like REST API if all it's needed is to get content from other sites.

tlovett1:
Awesome start, @kierahowe! Definitely agree with @nicholasio there are use cases where this is necessary and we should provide some best practices around those. I'd also love to see some info on why this isn't performant.

kierahowe:
Okay... Sounds good.   I will rework this a bit to a "recommend not using it" with a couple of alternatives and details why it isn't performant.

timwright12:
@kierahowe @nicholasio Any update here?

nicholasio:
@timwright12 no but if you assign this one to me I can work on it.

PR #200, opened 2017-09-20
Using the filesystem as storage can be troublesome

tott:
Given that many providers block filesystem writes or reads outside of the uploads folder for security reasons and/or run multi-server setups where filesystem writes would only go to one of the many servers the filesystem should not be considered as a data storage endpoint.


ericmann (approved):

tlovett1:
Awesome. Thanks @tott !

PR #205, opened 2017-11-09
Updates to options table/transients best practices

bengreeley:
Add updated information on the options table limits and issues when an options table is over 1MB in size (based off of Zach's blog post from March https://10up.com/blog/2017/wp-options-table/ ). Update transients API section to mention issues with large options tables.


lkwdwrd (approved):
This could definitely be expanded on a bit to explain how autoload and caching interact with each other. However, it's a good change even as is. We're not really trying to teach WP, simply talk through the best practices, which this does.

tlovett1:
Solid.

Issue #170, opened 2016-11-18
Caching WP_Query objects is a bad idea


pj-dave:
Hello 10up! In the [caching section](https://10up.github.io/Engineering-Best-Practices/php/#caching) you have several examples where you're saving WP_Query objects to the cache. We wrote a pretty detailed article about [why this is a bad idea](https://pressjitsu.com/blog/dont-cache-wp_query/), and why caching the complete HTML output or post IDs is a much better option.

tlovett1:
Totally agree. Would love to see a PR fixing this.

PR #212, opened 2018-01-17
Avoid caching WP Query object

nprasath002:
Fixes https://github.com/10up/Engineering-Best-Practices/issues/170

I modified the example code to save posts generated from WP Query instead on the query object itself


adamsilverstein:
Looks good to me, thanks @nprasath002!

PR #222, opened 2018-04-06
Updates for the Internationalization section

eugene-manuilov:


dana-ross:
The section on punctuation is useful and something I've seen quite often in my career. However, I think this is more general advice about internationalization and not specific to how we engineer for internationalization at 10up. I think this would be handled better with a link to a book or blog post like Otto's that engineers can refer to for more information.

@tlovett1 I'd like your opinion on commit #d7eba95 since that's a broader change in approach. It's something that's been on my mind as well since we so rarely take advantage of the translation functionality we build into everything. It's almost second nature for a lot of people to use `__()`, `_x()`, and their kin as they type code so we're not talking about a ton of lost productivity or anything by including it. But since we're not actually using the translation strings nobody is subjecting them to any kind of QC so we don't know if they're really any good.

tlovett1:
Thanks for putting this together, @eugene-manuilov !

I definitely like the puncatation part. Maybe separate that into another PR?

As for do not internationalize by default, I see your point and agree that it's probably not valuable in certain situations. However, it certainly depends. I would say making a front-end multilingual requires a lot of strategy. Whereas for an admin only plugin, it should just work as an admin can change their account language.

I'm in the camp of just internationalize by default is it doesn't add any real overhead in terms of performance.

Would love to hear other opinions.

dana-ross:
I believe internationalization _should_ be our default, and as much a cornerstone of 10up's development culture as early sanitization and late escaping. Having it in the code we write from day one buys us flexibility in future phases.

We also need to accept that if we're not explicitly supporting internationalization as part of a project, and we later need to add translation support, we will almost certainly need to revisit _some_ of the decisions we made. The original engineers on a project may have settled for `__()` and `_e()` everywhere, not even thinking about plurals or other concerns. But in that case, at least those function calls get us part of the way there and serve as landmarks in the source files.

eugene-manuilov:
Yes, I know that if we don't use i18n functions, we won't win too much in performance, but teaching developers to think before adding i18n and sanitization functions, we teach them to use lean development principles and practices, thus making better software ("the devil is in the details"). I can update my PR to better explain when we need to translate things and when we don't. Will it work?

P.S.: see [this question](https://stackoverflow.com/questions/3691625/why-are-php-function-calls-so-expensive) on the StackOverflow about why PHP functions are so expensive. Each i18n function calls a dozen of functions internally.

tlovett1:
Appreciate the PR and thoughtfulness, @eugene-manuilov. I think for now the consensus is we stick with the current approach.

PR #225, opened 2018-04-26
Remove Bower from recommended Tools

dana-ross:
Follow up on the abandoned PR #173. Just removes Bower from the recommended Tools. We can discuss replacing it with something else at another time.

@timwright12 does this work for you?


timwright12:
@daveross yeah, that's good for me, we're replacing it with NPM

PR #231, opened 2018-05-17
Added React Documentation

timwright12:
Added best practice documentation for React and Vue per the FE Team internal exercises.


_includes/markdown/React.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Figuring Out If React Is Right For Your Project
+
+When deciding whether React is the right tool for your project, it might help to ask the following questions:
+
+* Will this project involve building a user interface that requires statefulness? In other words, will your user interface need to...react...in real time to user actions or changeable data coming from an API?
+* Does your project require greater flexibility than a full-fledged MVC or MV(star) framework would allow? For example: does your application need to exist within a larger site context, rather than forming the basis for the entire site?
+* Does your project need enough reusable components to make it worth using React? React is a fairly heavy library compared with the similar Vue.
+* Is performance a primary concern? Will your project need to render large amounts of data in real time while remaining performant for the browser?
```
nicholasio: Are we providing an answer to this question? If the answer is yes are we gonna recommend React or not?

_includes/markdown/React.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Figuring Out If React Is Right For Your Project
+
+When deciding whether React is the right tool for your project, it might help to ask the following questions:
+
+* Will this project involve building a user interface that requires statefulness? In other words, will your user interface need to...react...in real time to user actions or changeable data coming from an API?
+* Does your project require greater flexibility than a full-fledged MVC or MV(star) framework would allow? For example: does your application need to exist within a larger site context, rather than forming the basis for the entire site?
```
nicholasio: I think this needs more direction on where React is the best choice or not based on the answer to the question.

tlovett1: I'm also unsure about this question. There have been times where I felt like React was a great choice even though I was only building one section of a website, maybe something like a plugin dashboard in the admin.

_includes/markdown/React.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Including with NPM
+
+If you already have a Node / NPM / Yarn project that you wish to include React as a dependency of, then you can use the installation instructions found on the [Official React Documentation](https://reactjs.org/docs/add-react-to-an-existing-app.html). Include React and ReactDOM in your package file just like you would any other dependency. Also be sure your transpiler is working to correctly process the ES6 and JSX from React into something the browser can understand.
+
+### Installation with a CLI
+
+[The React CLI](https://reactjs.org/blog/2016/07/22/create-apps-with-no-configuration.html) (create-react-app) provides us a quick way to setup and scaffold a React project. It makes use of the modern frontend workflows and out of the box, React CLI provides hot reloading, minification, asset management, module bundling, linting and a development server to test your application under realistic circumstances. It works with zero configuration from your part as everything is already set up.
+
+Using the CLI tool is ideal if you are creating a full React App with an API backend.
```
tlovett1, line 34: Would this be integrated into a WordPress theme?

_includes/markdown/React.md
```diff
@@ -0,0 +1,130 @@
...
+
+Redux is based almost completely on 3 main principles:
+
+1. Single Source of Truth: This is a main tree object which holds the entire state of your application. It’s meant to help your application to be seen as a whole.
+2. State is Read Only: The state of the application stops being mutable, the only way to change the state is by triggering an action, which itself will work as a log of what, when, and why changed in the state.
+3. Changes are made with pure functions: This means that only an action (or an event) can trigger a Reducer which will take the previous state and return a new one.
+
+### When to use Redux
+
+As appealing as it might be. Redux is not a tool for everyday use. By design, Redux will put constraints in your application that may not actually be needed. A good starting point to make this choice is the article by its creator, Dan Abramov: You might not need Redux. The usual recommendation is, think in React. And if along the way you discover the need of Redux, implement it.
```
tlovett1: Link?

tlovett1:
@timwright12 I made it through the React section. Can we break Vue out into a separate PR? That way it won't block us getting React merged.

_includes/markdown/React.md
timwright12: Doubtful, most likely just consuming the API like we did on Uber

_includes/markdown/React.md
timwright12: The intent here is, if you're answering "yes" to any of the questions, then React might be a good fit

_includes/markdown/React.md
timwright12: The biggest thing in using React right now is the weight of the library. So bringing it in for a small component on a site probably isn't the right way to go if we can accomplish the same result with something smaller like Vue. I agree that this question is a little confusing though. Maybe we can just lose it since it doesn't seem to help much.

_includes/markdown/React.md
timwright12: Will add

timwright12:
@tlovett1 Yeah, I can break out the Vue stuff

_includes/markdown/React.md
timwright12: Could be used for a plugin though

_includes/markdown/React.md
nicholasio: Maybe make that more clear? I see that in the Vue JS you explicitly say this, but I don't see this on the React section.

timwright12: Nice catch! Added

timwright12:
@tlovett1 Removed the Vue stuff, will submit a new PR for it

tlovett1 (approved):

PR #232, opened 2018-05-31
Updated package management documentation

timwright12:
Updated NPM with documentation instructions per our internal conversation.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    activesupport (4.2.9)
+    activesupport (4.2.10)
       i18n (~> 0.7)
       minitest (~> 5.1)
       thread_safe (~> 0.3, >= 0.3.4)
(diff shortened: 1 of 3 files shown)


dana-ross:
I know we've discussed our differences here, but this is from the co-founder & COO of npm, Inc.

> npm is always lowercase. The tool is not called NPM. The company is not called NPM, Inc.. It's always just npm.

https://twitter.com/seldo/status/850042251635867648?s=20

timwright12:
:groan:

timwright12:
@daveross lowercased npm throughout the docs

dana-ross:
Wouldn't we also use npm for managing dependencies for the rare node.js project?

nicholasio:
@daveross  maybe say we use npm for dependencies (without explicitly saying it's for frontend dependencies).

timwright12:
@nicholasio there's a little toe-stepping with Composer if we don't specify something IMO. How about "relevant dependencies"? @daveross That should cover, JS, non-JS, testing, and Node stuff.

timwright12:
Updated

dana-ross:
👍

PR #233, opened 2018-06-01
Added Vue Documentation

timwright12:


_includes/markdown/JavaScript.md
```diff
@@ -200,5 +200,7 @@ There are many JavaScript libraries available today. Many of them directly compe
 
 ### Frameworks
 
+[Vue](https://vuejs.org/) - Implementing Vue on a project allows us to take advantage of the statefulness built into something like React, but apply it on a much more lightweight and smaller scale as to not bog down performance by loading in a heavy library. [Learn more about how we use Vue]({{ site.baseurl }}/vue).
```
tlovett1: Are we saying Vue is a better tool than React at a smaller scale because of library size?

_includes/markdown/Vue.md
```diff
@@ -0,0 +1,110 @@
+The purpose of this document is to help guide you through working with a potential project using the JavaScript framework, Vue. Outside of The [Official Vue Documentation](https://vuejs.org/), you should find here personalized recommendations for using the framework based on the types of projects we typically see at 10up. We have broken these recommendations out into common elements we tend to interact with often in client work. If something you’re looking for isn’t represented you can either submit a pull request to this repository or refer to the [Official Vue Documentation](https://vuejs.org/v2/guide/).
+
+For general coding style recommendations please see the official [Vue style guide](https://vuejs.org/v2/style-guide/).
```
tlovett1: Shouldn't we just follow our general JS best practices?

_includes/markdown/Vue.md
```diff
@@ -0,0 +1,110 @@
...
+
+Using the right tool for the job is critical in navigating a successful project. Will Vue always be the answer? Of course not. But there are some instances where you may want to use this framework over something like (for example) a collection of plugins, custom JavaScript, or another framework.
+
+Some questions you may want to ask yourself:
+* Will this project contain a lot of components that need to interact with an API?
+* Are you replicating a lot features already present in Vue like client side templating and state management?
+* Will the addition of a framework save time and performance versus creating a custom solution? 
+* Do your client side templates need to be accessible to a back end technology (like PHP in a WordPress template)?
+
+If you find yourself answering, “Yes,” to those questions, Vue may be a great fit for your project! If not, a more traditional WordPress theme build may be the best.
```
tlovett1: This slightly implies to me that it's one or the other: WP theme or Vue.

_includes/markdown/Vue.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Implementing Vue on a Project
+
+There are several ways to include Vue.js in your project. From the inclusion of [Vue in your HTML page](https://vuejs.org/v2/guide/installation.html#CDN) via a script tag, to [installing via NPM](https://vuejs.org/v2/guide/installation.html#NPM) and through the [Vue CLI](https://vuejs.org/v2/guide/installation.html#CLI). We will take a look at the different methods and explain when to use them.
+
+### Script Element
```
tlovett1, line 26: I'm not sure how I feel about using the CDN. Isn't this just another DNS lookup to add to the list?

_includes/markdown/Vue.md
```diff
@@ -0,0 +1,110 @@
...
+
+If you are using Vue for isolated components within a single WordPress template be sure to enqueue the script as you normally would for only that instance as to not load Vue where you don’t need it. If you are using Vue throughout a project it is best to concatenate it alongside all your other JavaScript files.
+
+### Including with NPM
+
+[NPM](https://vuejs.org/v2/guide/installation.html#NPM) is the recommended installation method when building large scale applications with Vue. You can tweak your settings to either use Webpack or Browserify for module bundling. This method of installation is recommended for Vue-based applications when you want more custom and fine-grained control over the dependencies being used. You might also consider using this method when a project is taking advantage of NPM as a front-end package manager.
```
tlovett1: Is it worth mentioning Browserify?

_includes/markdown/Vue.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Installing with a CLI
+
+The [Vue CLI](https://vuejs.org/v2/guide/installation.html#CLI) provides us a quick way to setup and scaffold a Vue project. It makes use of the modern frontend workflows and out of the box, vue-CLI provides hot reloading, minification, asset management, module bundling, linting and a development server to test your application under realistic circumstances. It works with zero configuration from your part as everything is already set up in the [vue-CLI templates](https://github.com/vuejs-templates/). 
+
+The CLI caters from the most simple app ([the simple template](https://github.com/vuejs-templates/simple)), to the most complex one ([Webpack](https://github.com/vuejs-templates/webpack) or [Browserify](https://github.com/vuejs-templates/browserify) template). This is the recommended way to install Vue when you’re building a single page application that runs off an API and will be using the Vue framework exclusively. The CLI will likely need to be used and implemented at the beginning of a project.
```
tlovett1: I think we need to be more clear that this wouldn't be a WordPress theme.

_includes/markdown/Vue.md
```diff
@@ -0,0 +1,110 @@
...
+
+If you are integrating your templating with WordPress templates and back end engineering, we recommend using a script block that is embedded into the HTML template so back and front end engineers can use the output as a point of reference and easily work on the project within their discipline. [Read more about inline templates](https://vuejs.org/v2/guide/components.html#Inline-Templates).
+
+However, if you are building a single page application with Vue, it is recommended that each project module be componentize within a `.vue` template file. All JavaScript, CSS, and HTML templates would be scoped together.  For this, you will also likely use the [Vue CLI](https://vuejs.org/v2/guide/installation.html#CLI) to initialize your project. These are called “[Single File Components](https://vuejs.org/v2/guide/single-file-components.html).”
+
+If you are using Vue outside of the SPA context make sure all your features have non-JS fallback behaviors so nothing appears broken if JavaScript fails to load.
```
tlovett1: Are we still accounting for when a browser doesn't have JS turned on?

_includes/markdown/JavaScript.md
timwright12: Yeah. For basically one-off things. You don't need Webpack and misc. built tools for it either, so it's a lot easier to integrate within a project.

_includes/markdown/Vue.md
timwright12: This is more Vue-specific stuff, it doesn't interfere with our best practices and formatting

_includes/markdown/Vue.md
timwright12: Yeah, but it it really isn't any different than enqueuing a script. Ideally we'll include the script with NPM, but on the off case we can't I think using the CDN is fine (at least for the purpose of this documentation). I think there are enough cases where we'll use Vue for one specific feature where it doesn't need to be rolled into the main JS entry point that it's worth mentioning this as an option.

_includes/markdown/Vue.md
timwright12: Probably not, we don't use it. I can take it out.

_includes/markdown/Vue.md
timwright12: sure

_includes/markdown/Vue.md
timwright12: It's case by case, but overall, yes. Especially in this context.

_includes/markdown/Vue.md
timwright12: Updated this a little to be more clear

_includes/markdown/Vue.md
timwright12: Updated this

_includes/markdown/Vue.md
timwright12: Removed from 2 places

_includes/markdown/Vue.md
timwright12: Added a note about this not being a WP theme

timwright12:
@tlovett1 Updated based on feedback

_includes/markdown/Vue.md
```diff
@@ -0,0 +1,76 @@
+The purpose of this document is to help guide you through working with a potential project using the JavaScript framework, Vue. Outside of The [Official Vue Documentation](https://vuejs.org/), you should find here personalized recommendations for using the framework based on the types of projects we typically see at 10up. We have broken these recommendations out into common elements we tend to interact with often in client work. If something you’re looking for isn’t represented you can either submit a pull request to this repository or refer to the [Official Vue Documentation](https://vuejs.org/v2/guide/).
+
+For Vue-specific coding style and pattern recommendations please see the official [Vue style guide](https://vuejs.org/v2/style-guide/).
+
+## Figuring Out If Vue Is Right For Your Project
+
+Using the right tool for the job is critical in navigating a successful project. Will Vue always be the answer? Of course not. But there are some instances where you may want to use this framework over something like (for example) a collection of plugins, custom JavaScript, or another framework.
```
tlovett1: @timwright12 I think your paragraph about Vue vs. React should be up here.

timwright12:
@tlovett1 added:

> "For instance, if you're building a larger-scale application that you feel needs a stateful system, we've found that React is a often a better option when compared to Vue."

How's that?

_includes/markdown/Vue.md
```diff
@@ -0,0 +1,76 @@
...
+
+Some questions you may want to ask yourself:
+* Will this project contain a lot of components that need to interact with an API?
+* Are you replicating a lot features already present in Vue like client side templating and state management?
+* Will the addition of a framework save time and performance versus creating a custom solution? 
+* Do your client side templates need to be accessible to a back end technology (like PHP in a WordPress template)?
+
+If you find yourself answering, “Yes,” to those questions, Vue may be a great fit for your project or component! If not, a more traditional build may be the best.
+
+If you're answering "no" to some of those questions, but still feel a JavaScript-framework solution is the correct model, you may want to read through our [React]({{ site.baseurl }}/react) recommendations as a path forward for a larger scale application.
```
tlovett1: @timwright12 seems like the only question you might answer no to that would push you towards React is the last one?

timwright12: @tlovett1 Yeah, I would expect a good amount of overlap since they're both pretty similar. The main pieces are the templating style you need and the project/task size that are the differentiators between the two paths.

_includes/markdown/Vue.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Figuring Out If Vue Is Right For Your Project
+
+Using the right tool for the job is critical in navigating a successful project. Will Vue always be the answer? Of course not. But there are some instances where you may want to use this framework over something like (for example) a collection of plugins, custom JavaScript, or another framework. For instance, if you're building a larger-scale application that you feel needs a stateful system, we've found that [React]({{ site.baseurl }}/react) is a often a better option when compared to Vue.
+
+Some questions you may want to ask yourself:
+* Will this project contain a lot of components that need to interact with an API?
+* Are you replicating a lot features already present in Vue like client side templating and state management?
+* Will the addition of a framework save time and performance versus creating a custom solution? 
+* Do your client side templates need to be accessible to a back end technology (like PHP in a WordPress template)?
```
saltcod: I feel like there's another use case here that isn't covered by these differentiators. Beyond small, reusable "components", Vue is also an excellent choice for larger features on a site, eg:  a content filtering / search interface, image gallery filtering experience, map filtering / searching etc.  

These are examples of projects that I wouldn't necessarily want to use React for, nor would I want to do with vanilla js (because Vue gives you so much of what you need here for free). 

tl;dr:

In light of this, I propose a new bullet point here: 

* Are you creating a relatively large, stand-alone feature in your project (e.g.: a stateful interface for filtering content)

saltcod: @timwright12 @tlovett1 ^

_includes/markdown/Vue.md
```diff
@@ -0,0 +1,76 @@
...
+
+If you find yourself answering, “Yes,” to those questions, Vue may be a great fit for your project or component! If not, a more traditional build may be the best.
+
+If you're answering "no" to some of those questions, but still feel a JavaScript-framework solution is the correct model, you may want to read through our [React]({{ site.baseurl }}/react) recommendations as a path forward for a larger scale application. 
+
+## Implementing Vue on a Project
```
saltcod, line 22: After this page is published I'd like to revisit and some specifics of implementing Vue in a WP project. 

As mentioned, I feel like Vue is an excellent tool for medium-large sized features within a site, but using Vue within a WordPress environment has some nuance. 

Examples: 
What is a "medium-large" feature — we could include a few examples of search/filtering interfaces 
How do we recommend handling api queries?  
How do we handle sharing template-part partials between WordPress and Vue? 
etc.

I haven't read much out there about good ways of thinking about these WordPress + Vue problems, so I think our voice here would be a welcome one.  

Again, I wouldn't want this to hang up the PR process here, but rather supplement the content when this is published.

_includes/markdown/Vue.md
timwright12: @saltcod When we did our Vue team learning a few months ago, we settled on using React for large things and Vue for small things. I would actually still consider everything you listed there to be relatively small features in the grand scheme of a project.

I also don't want to fall back on statefulness as a qualifier to use a JS framework as every project we do contains some for of statefulness. This is something that needs to be updated in the React doc as well.

_includes/markdown/Vue.md
timwright12: @saltcod I'll flag this for discussion after initial merge.

_includes/markdown/Vue.md
saltcod: @timwright12 Yep agreed.  When I say relatively large features I'm suggesting that there's a tipping point between Vue and vanilla, not Vue and React.

_includes/markdown/JavaScript.md
```diff
@@ -143,4 +143,6 @@ There are many JavaScript libraries available today. Many of them directly compe
 
 [React](https://reactjs.org/) - Using React provides a library to create large-scale, stateful JavaScript applications. It aims to provide a flexible system of creating highly componentized user interfaces. [Learn more about how we use React]({{ site.baseurl }}/react).
 
+[Vue](https://vuejs.org/) - Implementing Vue on a project allows us to take advantage of the statefulness built into something like React, but apply it on a much more lightweight and smaller scale as to not bog down performance by loading in a heavy library. [Learn more about how we use Vue]({{ site.baseurl }}/vue).
```
tlovett1, line 184: How "heavy" is React? Is this really a major concern today?

_includes/markdown/Vue.md
```diff
@@ -0,0 +1,76 @@
+The purpose of this document is to help guide you through working with a potential project using the JavaScript framework, Vue. Outside of The [Official Vue Documentation](https://vuejs.org/), you should find here personalized recommendations for using the framework based on the types of projects we typically see at 10up. We have broken these recommendations out into common elements we tend to interact with often in client work. If something you’re looking for isn’t represented you can either submit a pull request to this repository or refer to the [Official Vue Documentation](https://vuejs.org/v2/guide/).
+
+For Vue-specific coding style and pattern recommendations please see the official [Vue style guide](https://vuejs.org/v2/style-guide/).
+
+## Figuring Out If Vue Is Right For Your Project
```
tlovett1, line 5: @timwright12 @saltcod I think we need to redo this section.

I am still unclear on the React versus Vue piece. The sentence: "If you're answering "no" to some of those questions, but still feel a JavaScript-framework solution is the correct model, you may want to read through our [React]({{ site.baseurl }}/react) recommendations as a path forward for a larger scale application." makes very little sense to me. If my project doesn't contain a lot of components that need to interact with an API, React is a better fit?

Does React allow for client side templating at all? Seems like the reasons to use Vue are client side templating and lighter library.

_includes/markdown/JavaScript.md
timwright12: It's pretty heavy, you also need React DOM and either a full webpack set up or a JSX client-side converter. So yeah, it's a pretty significant thing to add into a project when held up against Vue where you don't need anything other than the library itself.

_includes/markdown/Vue.md
timwright12: @tlovett1 Terry ( @saltcod ) and I are going to take this offline and rework the section. It shouldn't be a this vs. that kind of thing. They should be independently evaluated for the benefit of the project. Ultimately before you even use one you need to answer the question: "Do I need a JS framework at all" - so those parts will be the same in each document. Beyond that there should just be a couple things that nudge you towards one and away from the other. But I want to get away from the angle/perception that the presence of a Vue doc is anti-React and vice versa.

timwright12:
@tlovett1 Terry and I rewrote this documentation, please review when you can some time. (cc @saltcod)

tlovett1:
@timwright12 I'm good with this getting merged. Nice job @timwright12  and @saltcod !

PR #154, opened 2016-10-03
Feature add skip links accessibility

RachelRVasquez:

PR #234, opened 2018-06-12
Add a section about bypass blocks (skip links) for a11y

pattonwebz:
This is a refresh of #154

### DIFF
--- a/_includes/markdown/Markup.md
+++ b/_includes/markdown/Markup.md
@@ -187,17 +187,6 @@ For design projects and projects with a global marketplace (companies with entit
 
 While [Section 508](https://www.section508.gov/) is the US standard, following the guidance of WCAG 2.0 will help a project pass Section 508 and also maintain a consistent internal standard. If a project specifically requires Section 508, additional confirmation testing can be done.
 
-### ARIA Landmark Roles
-ARIA (Assistive Rich Internet Applications) is a spec from the W3C. It was created to improve accessibility of applications by providing extra information to screen readers via HTML attributes. Screen readers can already read HTML, but ARIA can help add context to content and make it easier for screen readers to interact with content.
-
-ARIA is a descriptive layer on top of HTML to be used by screen readers. It has no effect on how elements are displayed or behave in browsers. We use these ARIA Landmark Roles (banner, navigation, main, etc.) to provide a better experience to users with disabilities.
-
-Example:
-
-```html
-<header id="masthead" class="site-header" role="banner"></header>
-```
-
 ### States and Properties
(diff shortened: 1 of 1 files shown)


_includes/markdown/Markup.md
```diff
@@ -233,6 +233,26 @@ Form elements should also be logically grouped using the ```<fieldset>``` elemen
 
 Finally, we should ensure that all interactive elements are keyboard navigable, providing easy use for people with vision or mobility disabilities (or those not able to use a mouse). In general, the tab order should be dictated by a logical source order of elements. If you feel the need to change the tab order of certain elements, it likely indicates that you should rework the markup to flow in a logical order or have a conversation with the designer about updates that can be made to the layout to improve accessibility.
 
+### Bypass Blocks
+
+[Bypass blocks](https://developer.wordpress.org/themes/functionality/accessibility/#skip-links), sometimes knows as 'skip to content links' - or simply 'skip links' - are links in a document to allow users that rely on screen readers, keyboard navigation or other assistive technologies to _bypass_ certain page elements or _skip_ to a specific section of a page with ease.
```
timwright12: @pattonwebz Bypass are more than just skip links, it's also aria-roles and some other method of chunking parts of a document for assistive technology to use. Here's some more information: https://www.w3.org/WAI/WCAG21/quickref/?versions=2.0#bypass-blocks

_includes/markdown/Markup.md
```diff
@@ -233,6 +233,26 @@ Form elements should also be logically grouped using the ```<fieldset>``` elemen
...
+
+An example of what a skip link might look like in a template file:
+
+```html
+<a class="skip-link screen-reader-text" href="#content">
+  <?php esc_html_e( 'Skip to content', 'my-domain' ); ?>
+</a>
+```
+
+Bypass blocks make use of CSS to hide them from sighted users and keep them accessible for screen readers. Usually the styles are attached to a ```screen-reader-text``` class of some kind. This CSS is used to position the links off the screen then use ```:focus``` styles to make the link visible for sighted keyboard users.
```
timwright12: @pattonwebz This would apply for skip links, but not all bypass block. We might need some sub sections in here to account for everything.

timwright12 (changes requested):
If you don't have time to make these updates, let me know and I can take over the branch.

_includes/markdown/Markup.md
pattonwebz: Agreed, some sub sections will be good. I struggled to make this all fit into just a few paragraphs.

Will take another pass at this over the weekend see what I come up with :+1:

timwright12: @pattonwebz any progress here?

timwright12 (approved):

Issue #245, opened 2018-08-07
Remove sanitize.css and normalize.css as primary tool.


smy315:
Since both Theme-Scaffold and Plugin-Scaffold uses normalize now (https://github.com/10up/plugin-scaffold/issues/37 and https://github.com/10up/theme-scaffold/issues/45) for CSS reset. Also, since sanitize is no longer a 10up project? 

PR: https://github.com/10up/Engineering-Best-Practices/pull/246

PR #246, opened 2018-08-07
Removed sanitize.css and added normalize.css as primary tool for CSS reset

smy315:
https://github.com/10up/Engineering-Best-Practices/issues/245


jonathantneal:
Good catch. I never noticed this was still in the repo. @timwright12, since you use babel and postcss and include a browserslistrc file, I recommend looking at the browserslist powered version of normalize:

https://github.com/csstools/postcss-normalize

PR #251, opened 2018-08-24
JS / Client-side Data (Ajax) Best Practices

timwright12:
Added part 1 of the FE team's collaborative document on client-side data


_includes/markdown/JavaScript.md
```diff
@@ -67,6 +67,44 @@ document.getElementById( 'menu' ).addEventListener( 'click', ( e ) => {
 ```
 You may be wondering why we don't just add one listener to the ```<body>``` for all our events. Well, we want the event to *bubble up the DOM as little as possible* for [performance reasons](https://jsperf.com/event-delegation-distance). This would also be pretty messy code to write.
 
+<h2 id="client-side-data" class="anchor-heading">Client-side Data {% include Util/top %}</h2>
+
+When dealing with client-side data requests (Ajax calls), there are a lot of different methods to consider. This portion of the document will walk you through various situations and talk about the different technologies and patterns you may encounter along the way.
+
+### Using Fetch and Promises for Modern Environments
+The Fetch API is a modern replacement for the XMLHttpRequest. It is [generally well supported](https://caniuse.com/#search=fetch), having features present in all evergreen browsers (browsers that auto-update). Fetch is recommended to be used in all modern environments when making Ajax calls or dealing with client-side data requests. Visit the [MDN Fetch documentation](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) for a basic example of how to use this API.
```
ericmann, line 75: IE11 support is still a requirement for many enterprise installations (though Edge is a superior browser). Recommend a preferred IE polyfill to ensure proper compatibility.

ericmann: Nevermind, saw a note a couple of paragraphs below. Might be worth a parenthetical here as well, though, that references the to-be-mentioned polyfill below?

PR #252, opened 2018-08-30
Update list of local environments

bengreeley:
Updated the list of local environments WP-CLI is included on by default to encourage engineers to use WP Local Docker.


dana-ross:
@tlovett1 Should we remove the mention of VIP Quickstart while we're at it?

tlovett1:
Thanks @bengreeley and that would be great @daveross

bengreeley:
Good call @daveross - I removed VIP Quickstart

PR #256, opened 2018-10-03
Add a section about avoiding XSS in JavaScript

dkoo:
Hat tip to @smy315 for the initial writeup and solution examples. Given the recent conversation around this (and running into it in various projects), I thought XSS warranted a section in the JavaScript best practices.

### DIFF
--- a/_includes/markdown/JavaScript.md
+++ b/_includes/markdown/JavaScript.md
@@ -165,7 +165,7 @@ const myElement = document.getElementById( 'myElement' );
 myElement.innerHTML = `<div class="container"><a href="${ someUrl }">${ someContent }</a></div>`;
 ```
 
-However, passing HTML strings to ```innerHTML``` and methods like it (including jQuery's commonly used ```append``` and ```html``` methods) can expose your code to cross-site scripting, also known as XSS—the most common security vulnerability in JavaScript. Because these methods evaluate strings passed to them as HTML, they can execute potentially harmful code. For example, what happens if ```someContent``` in the above example is ```<img src="fakeImage" onerror="alert( 'hacked' ) />```? The JavaScript in the ```onerror``` attribute is executed.
+However, passing HTML strings to ```innerHTML``` and methods like it (including jQuery's commonly used ```append``` and ```html``` methods) can expose your code to cross-site scripting, also known as XSS—the most common security vulnerability in JavaScript. Because these methods evaluate strings passed to them as HTML, they can execute potentially harmful code. For example, what happens if ```someContent``` in the above example is ```<img src="fakeImage" onerror="alert( 'hacked' )" />```? The JavaScript in the ```onerror``` attribute is executed.
 
 There are several measures you can take to circumvent this XSS vulnerability:
 
--- a/_includes/markdown/JavaScript.md
+++ b/_includes/markdown/JavaScript.md
@@ -157,12 +157,14 @@ Some older projects that have not yet been upgraded may not have the capability 
 
 ### Secure Your Code
 
-In JavaScript, we often have to insert new elements with dynamic attributes and content into the DOM. A common way to do this is to use the ```innerHTML``` method like so:
+In JavaScript, we often have to insert new elements with dynamic attributes and content into the DOM. A common way to do this is to use the [```innerHTML```](https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML) method like so:
 
 ```javascript
-const myElement = document.getElementById( 'myElement' );
+const someElement = document.getElementById( 'someElement' );
+const someUrl = 'https://someurl.com/';
+const someContent = 'Some content';
 
-myElement.innerHTML = `<div class="container"><a href="${ someUrl }">${ someContent }</a></div>`;
+someElement.innerHTML = `<div class="container"><a href="${ someUrl }">${ someContent }</a></div>`;
 ```
 
 However, passing HTML strings to ```innerHTML``` and methods like it (including jQuery's commonly used ```append``` and ```html``` methods) can expose your code to cross-site scripting, also known as XSS—the most common security vulnerability in JavaScript. Because these methods evaluate strings passed to them as HTML, they can execute potentially harmful code. For example, what happens if ```someContent``` in the above example is ```<img src="fakeImage" onerror="alert( 'hacked' )" />```? The JavaScript in the ```onerror``` attribute is executed.
@@ -171,20 +173,22 @@ There are several measures you can take to circumvent this XSS vulnerability:
 
 #### Use ```textContent``` instead of ```innerHTML```
 
-When setting the human-readable content of a single element, using ```textContent``` is safer than using ```innerHTML``` because it does not parse strings as HTML—meaning any malicious code passed to it will not be executed.
+When setting the human-readable content of a single element, using [```textContent```](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent) is safer than using ```innerHTML``` because it does not parse strings as HTML—meaning any malicious code passed to it will not be executed.
 
 ```javascript
-const myElement = document.getElementById( 'myElement' );
+const anotherElement = document.getElementById( 'anotherElement' );
 
-myElement.textContent = 'This will get displayed as human-readable text, even if it contains HTML-like <syntax>.'
+anotherElement.textContent = 'This will get displayed as human-readable text, even if it contains HTML-like <syntax>.';
 ```
 
 #### Use the ```Element``` API instead of HTML-parsing methods
 
-When you need to create multiple DOM elements, use JavaScript's [```Element``` API](https://developer.mozilla.org/en-US/docs/Web/API/Element) to create elements, set attributes and append them to the document. Creating your own elements and attributes will ensure that only those you explicitly define will make their way into the DOM. Refactoring our first example:
+When you need to create multiple DOM elements, use JavaScript's [```Element```](https://developer.mozilla.org/en-US/docs/Web/API/Element) API to create elements, set attributes and append them to the document. Creating your own elements and attributes will ensure that only those you explicitly define will make their way into the DOM. Refactoring our first example:
 
 ```javascript
-const myElement = document.getElementById( 'myElement' );
+const someElement = document.getElementById( 'someElement' );
+const someUrl = 'https://someurl.com/';
+const someContent = 'Some content';
 
 // Create the new elements.
 const container = document.createElement( 'div' );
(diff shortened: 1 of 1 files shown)


_includes/markdown/JavaScript.md
```diff
@@ -155,6 +155,74 @@ On all new projects you should be using up to date JavaScript methodologies comb
 
 Some older projects that have not yet been upgraded may not have the capability to use the most modern techniques, but it is still important to have processes in place that allow us to grow the technology stack as a project matures. In these cases, you should still follow best practice recommendations even if the newest patterns are not yet available to you.
 
+### Secure Your Code
+
+In JavaScript, we often have to insert new elements with dynamic attributes and content into the DOM. A common way to do this is to use the ```innerHTML``` method like so:
```
JodiWarren: Maybe you could provide link to the MDN page here https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML

_includes/markdown/JavaScript.md
```diff
@@ -155,6 +155,74 @@ On all new projects you should be using up to date JavaScript methodologies comb
...
+myElement.innerHTML = `<div class="container"><a href="${ someUrl }">${ someContent }</a></div>`;
+```
+
+However, passing HTML strings to ```innerHTML``` and methods like it (including jQuery's commonly used ```append``` and ```html``` methods) can expose your code to cross-site scripting, also known as XSS—the most common security vulnerability in JavaScript. Because these methods evaluate strings passed to them as HTML, they can execute potentially harmful code. For example, what happens if ```someContent``` in the above example is ```<img src="fakeImage" onerror="alert( 'hacked' )" />```? The JavaScript in the ```onerror``` attribute is executed.
+
+There are several measures you can take to circumvent this XSS vulnerability:
+
+#### Use ```textContent``` instead of ```innerHTML```
+
+When setting the human-readable content of a single element, using ```textContent``` is safer than using ```innerHTML``` because it does not parse strings as HTML—meaning any malicious code passed to it will not be executed.
```
JodiWarren: Maybe make ```textContent``` link to the MDN page: https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent

_includes/markdown/JavaScript.md
```diff
@@ -155,6 +155,74 @@ On all new projects you should be using up to date JavaScript methodologies comb
...
+
+```javascript
+const myElement = document.getElementById( 'myElement' );
+
+myElement.textContent = 'This will get displayed as human-readable text, even if it contains HTML-like <syntax>.'
```
smy315: Semicolon missing.

_includes/markdown/JavaScript.md
```diff
@@ -155,6 +155,74 @@ On all new projects you should be using up to date JavaScript methodologies comb
...
+
+#### Sanitize HTML strings before appending to the DOM
+
+You'll notice that the refactored example is much more verbose than the original example—what was two lines of executable code became eight. This can become painful if you need to create many new elements. In those cases, the convenience of methods like ```innerHTML``` can be extremely tempting. 
+
+If you must use them, consider adding a client-side sanitization library to your project so you can strip potentially malicious code from your HTML before you add it to the DOM. Two popular libraries are [DOMPurify](https://github.com/cure53/DOMPurify) and [js-xss](https://github.com/leizongmin/js-xss). They work similarly:
+
+```javascript
+// Using DOMPurify
+DOMPurify.sanitize( '<a href="#" onclick="alert( 'Alert!' );">link</a>' ); // Becomes <a href="#">link</a>
```
JodiWarren: It might be nice to provide the same example as on line 163 here, just to keep the examples consistent.

_includes/markdown/JavaScript.md
```diff
@@ -155,6 +155,74 @@ On all new projects you should be using up to date JavaScript methodologies comb
...
+```javascript
+const myElement = document.getElementById( 'myElement' );
+
+// Create the new elements.
+const container = document.createElement( 'div' );
+const anchor = document.createElement( 'a' );
+
+// Set attributes and content.
+container.setAttribute( 'class', 'container' );
+anchor.setAttribute( 'href', someUrl );
```
smy315: Put someUrl in single quotes or assign a url to this variable.

_includes/markdown/JavaScript.md
```diff
@@ -155,6 +155,74 @@ On all new projects you should be using up to date JavaScript methodologies comb
...
+anchor.textContent = someContent;
```
smy315: Put someContent in single quotes.

smy315 (changes requested):
Looks good, left few minor comments on the examples.

_includes/markdown/JavaScript.md
```diff
@@ -155,6 +155,74 @@ On all new projects you should be using up to date JavaScript methodologies comb
 
 Some older projects that have not yet been upgraded may not have the capability to use the most modern techniques, but it is still important to have processes in place that allow us to grow the technology stack as a project matures. In these cases, you should still follow best practice recommendations even if the newest patterns are not yet available to you.
 
+### Secure Your Code
+
+In JavaScript, we often have to insert new elements with dynamic attributes and content into the DOM. A common way to do this is to use the ```innerHTML``` method like so:
+
+```javascript
+const myElement = document.getElementById( 'myElement' );
```
JodiWarren: Maybe change `myElement` to `element` to differentiate it from the elements that you're creating. It's has a pretty mild effect here, but in the more verbose `textContent` example it might add a little bit of clarity.

JodiWarren (approved):
Nice work dude! I made a few suggestions - but please feel free to take or ignore them!

_includes/markdown/JavaScript.md
```diff
@@ -155,6 +155,74 @@ On all new projects you should be using up to date JavaScript methodologies comb
...
+
+myElement.textContent = 'This will get displayed as human-readable text, even if it contains HTML-like <syntax>.'
+```
+
+#### Use the ```Element``` API instead of HTML-parsing methods
+
+When you need to create multiple DOM elements, use JavaScript's [```Element``` API](https://developer.mozilla.org/en-US/docs/Web/API/Element) to create elements, set attributes and append them to the document. Creating your own elements and attributes will ensure that only those you explicitly define will make their way into the DOM. Refactoring our first example:
```
smy315: Extra space linked before API.

nicholasio (commented):
This looks great to me. I'd add a note saying that including the library is only worth it if the code base is generating a lot of HTML dynamically. For some projects we have just a few spots whee JS is generating HTML that is generally better to just use the Element API IMO.

_includes/markdown/JavaScript.md
dkoo: Nice idea—I added it as the first example for both DOMPurify and js-xss. Also updated the output for js-xss to reflect what it actually was upon testing.

_includes/markdown/JavaScript.md
dkoo: I assigned it a value instead in all examples.

_includes/markdown/JavaScript.md
dkoo: I assigned it a value in all examples.

dkoo:
@JodiWarren Thanks for the suggestions! I took most of them.

dkoo:
@nicholasio I added a note in the beginning of the last subsection stating that the `Element` API is the preferred best practice, but that a sanitizer is another option for projects that need to generate a lot of dynamic HTML.

dkoo:
@smy315 Feedback addressed!

smy315 (approved):

_includes/markdown/JavaScript.md
```diff
@@ -155,6 +155,80 @@ On all new projects you should be using up to date JavaScript methodologies comb
...
+
+```javascript
+const someElement = document.getElementById( 'someElement' );
+const someUrl = 'https://someurl.com/';
+const someContent = 'Some content';
+
+someElement.innerHTML = `<div class="container"><a href="${ someUrl }">${ someContent }</a></div>`;
+```
+
+However, passing HTML strings to ```innerHTML``` and methods like it (including jQuery's commonly used ```append``` and ```html``` methods) can expose your code to cross-site scripting, also known as XSS—the most common security vulnerability in JavaScript. Because these methods evaluate strings passed to them as HTML, they can execute potentially harmful code. For example, what happens if ```someContent``` in the above example is ```<img src="fakeImage" onerror="alert( 'hacked' )" />```? The JavaScript in the ```onerror``` attribute is executed.
```
timwright12: @dkoo In this paragraph:

no need to reference jQuery as we're migrating away from it.
space between XSS and "the"?
remove "what happens if"

_includes/markdown/JavaScript.md
```diff
@@ -155,6 +155,80 @@ On all new projects you should be using up to date JavaScript methodologies comb
...
+
+There are several measures you can take to circumvent this XSS vulnerability:
+
+#### Use ```textContent``` instead of ```innerHTML```
+
+When setting the human-readable content of a single element, using [```textContent```](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent) is safer than using ```innerHTML``` because it does not parse strings as HTML—meaning any malicious code passed to it will not be executed.
+
+```javascript
```
timwright12: @dkoo can we replace this code block with a link to documentation about using `textContent`?

_includes/markdown/JavaScript.md
```diff
@@ -155,6 +155,80 @@ On all new projects you should be using up to date JavaScript methodologies comb
...
+const anotherElement = document.getElementById( 'anotherElement' );
+
+anotherElement.textContent = 'This will get displayed as human-readable text, even if it contains HTML-like <syntax>.';
+```
+
+#### Use the ```Element``` API instead of HTML-parsing methods
+
+When you need to create multiple DOM elements, use JavaScript's [```Element```](https://developer.mozilla.org/en-US/docs/Web/API/Element) API to create elements, set attributes and append them to the document. Creating your own elements and attributes will ensure that only those you explicitly define will make their way into the DOM. Refactoring our first example:
+
+```javascript
```
timwright12: @dkoo there's a ton of stuff in the Element API, rather than addressing a very small portion of it, let's just link off to MDN.

_includes/markdown/JavaScript.md
```diff
@@ -155,6 +155,80 @@ On all new projects you should be using up to date JavaScript methodologies comb
...
+
+// Append new elements to the DOM all at once.
+someElement.appendChild( container );
+```
+
+Note that appending new elements to the DOM is a relatively expensive operation, so generally you'll want to build the structure of new, nested elements _before_ adding them to the DOM, then append them to an existing element all at once by appending a single container element.
+
+#### Sanitize HTML strings before appending to the DOM
+
+In general, using the ```Element``` API is the preferred best practice to safely create and add DOM elements. However, you'll notice that the refactored example is much more verbose than the original example—what was just a few lines of executable code more than doubled after refactoring. This can become painful if you need to create many new elements. In those cases, the convenience of methods like ```innerHTML``` can be extremely tempting.
```
timwright12: @dkoo reword this a little if the code block above it gone

_includes/markdown/JavaScript.md
```diff
@@ -155,6 +155,80 @@ On all new projects you should be using up to date JavaScript methodologies comb
...
+
+Note that appending new elements to the DOM is a relatively expensive operation, so generally you'll want to build the structure of new, nested elements _before_ adding them to the DOM, then append them to an existing element all at once by appending a single container element.
+
+#### Sanitize HTML strings before appending to the DOM
+
+In general, using the ```Element``` API is the preferred best practice to safely create and add DOM elements. However, you'll notice that the refactored example is much more verbose than the original example—what was just a few lines of executable code more than doubled after refactoring. This can become painful if you need to create many new elements. In those cases, the convenience of methods like ```innerHTML``` can be extremely tempting. 
+
+If you need to generate a large amount of HTML dynamically, consider adding a client-side sanitization library to your project so you can strip potentially malicious code from your HTML before you add it to the DOM. Two popular libraries are [DOMPurify](https://github.com/cure53/DOMPurify) and [js-xss](https://github.com/leizongmin/js-xss). They work similarly:
+
+```javascript
```
timwright12: @dkoo If we want to sanction this library, that's fine with me, but I don't really want to have to maintain third party code blocks in our best practice documentation. Let's remove this block in favor of a link to documentation.

_includes/markdown/JavaScript.md
```diff
@@ -155,6 +155,80 @@ On all new projects you should be using up to date JavaScript methodologies comb
...
+// Using DOMPurify
+DOMPurify.sanitize( `<div class="container"><a href="https://someurl.com/"><img src="fakeImage" onerror="alert( 'hacked' )" /></a></div>` ); // Becomes <div class="container"><a href="https://someurl.com/"><img src="fakeImage"></a></div>
+DOMPurify.sanitize( '<a href="#" onclick="alert( 'Alert!' );">link</a>' ); // Becomes <a href="#">link</a>
+DOMPurify.sanitize( '<UL><li>Item</UL>' ); // Becomes <ul><li>Item</li></ul>
+DOMPurify.sanitize( '<svg><g/<script></svg>' ); // Becomes <svg><g></g></svg>
+
+// Using js-xss
+filterXSS( `<div class="container"><a href="https://someurl.com/"><img src="fakeImage" onerror="alert( 'hacked' )" /></a></div>` ); // Becomes <div><a href="https://someurl.com/"><img src /></a></div>
```
timwright12: @dkoo same comment about third party code

_includes/markdown/JavaScript.md
```diff
@@ -155,6 +155,80 @@ On all new projects you should be using up to date JavaScript methodologies comb
...
+// Create structure for the new elements.
+container.appendChild( anchor );
+
+// Append new elements to the DOM all at once.
+someElement.appendChild( container );
+```
+
+Note that appending new elements to the DOM is a relatively expensive operation, so generally you'll want to build the structure of new, nested elements _before_ adding them to the DOM, then append them to an existing element all at once by appending a single container element.
+
+#### Sanitize HTML strings before appending to the DOM
```
timwright12: @dkoo is it worth mentioning DOMParser here?

timwright12 (changes requested):

_includes/markdown/JavaScript.md
dkoo: Removed the jQuery reference and reworded the "what happens if". No space after "XSS"—using an em dash there!

dkoo:
@timwright12 Thanks for the feedback! I replaced most of the code blocks with links to the MDN documentation and removed references to the specific 3rd party libraries. Also added a paragraph about `DOMParser`, as you suggested.

_includes/markdown/JavaScript.md
```diff
@@ -155,6 +155,43 @@ On all new projects you should be using up to date JavaScript methodologies comb
...
+
+However, passing HTML strings to ```innerHTML``` and methods like it can expose your code to [cross-site scripting](https://developer.mozilla.org/en-US/docs/Glossary/Cross-site_scripting), also known as XSS—the most common security vulnerability in JavaScript. Because these methods evaluate strings passed to them as HTML, they can execute potentially harmful code. For instance, if ```someContent``` in the above example is ```<img src="fakeImage" onerror="alert( 'hacked!' )" />```, the JavaScript in the ```onerror``` attribute will be executed.
+
+There are several measures you can take to circumvent this XSS vulnerability:
+
+#### Use ```textContent``` instead of ```innerHTML```
+
+When setting the human-readable content of a single element, using ```textContent``` is safer than using ```innerHTML``` because it does not parse strings as HTML—meaning any malicious code passed to it will not be executed. Refer to [MDN's documentation on ```textContent```](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent) for more info.
+
+```
```
timwright12: @dkoo rogue dashes?

timwright12 (changes requested):

_includes/markdown/JavaScript.md
jonathantneal: You’ll need to remove this rogue line. It’s causing formatting issues.

dkoo:
@timwright12 @jonathantneal Whoops, rogue characters deleted!

timwright12 (approved):

jonathantneal:
I’m a _big_ fan of this new “Secure Your Code” section. Well done, @dkoo!

PR #257, opened 2018-10-04
Minor cleanup

dkoo:
This 1.) updates the TOC page to reflect the current JavaScript page structure, and 2.) adds a specific recommendation to use `document.createElement` for creating new elements in the "Secure your code" subsection.


jonathantneal:
My two cents: I think the heading could be kept short, as in:

```
Use the DOM API to create and add elements
```

dkoo:
@jonathantneal Great suggestion—I changed it to "Use the DOM API..." as both `document` and `Element` are subsets.

PR #258, opened 2018-10-04
Remove references to VVV from best practices.

bengreeley:
In an effort to standardize development workflows at 10up we should promote WP Local Docker as the officially supported development environment for easier integration with WP Snapshots. This removes references to VVV and solely refers to WP Local Docker.

Updated links to point to WP Local Docker docs site instead of the GitHub repository.


dana-ross:
"local development environment for WordPress development" sounds repetitive. What if you took out the first "development", making it a "local environment for WordPress development"?

tlovett1:
Looks good to me.

PR #259, opened 2018-10-05
Test Users on Production

tlovett1:
Added a note about accidentally importing test users. Would love any feedback.


dana-ross:
Rather than focus on test users, how about just a general non-subscriber user audit after every migration? I was just talking with Ryan A about Administrator user accounts created during UAT that need to be downgraded in production, or third parties given access during UAT who shouldn't be able to access the live site. This should be done in coordination with the client.

dkotter:
Yeah, I think this looks good but I think adding a line or two to the end of this, mentioning a user audit after migration is probably a good call.

Issue #244, opened 2018-08-07
Consider renaming the "Structure" page to "Design"


helen:
This is coming out of #229 to add a section about privacy best practices (specifically user data right now, but could be more things). It doesn't really fit anywhere else because it's not language-specific like a lot of other security/privacy information, and I think `Design` does a good job of describing both things like file structure and other general engineering considerations.

tlovett1:
I like this idea! Want to draft a PR?

helen:
@tlovett1 Yep, see #262

PR #229, opened 2018-05-09
Add privacy best practices section

adamsilverstein:
* Aims to guide developers on best practices around privacy.
* Project structure was the best place I could find for this information without creating a new section. Maybe we can change this section name to 'Project Practices' or something like that.

Fixes https://github.com/10up/Engineering-Best-Practices/issues/220

PR #262, opened 2018-10-23
Rename `(Project) Structure` to `(Project) Design`

helen:
See #244 

This also adds `Privacy` to the TOC per #229, it was easier to do it this way to avoid conflicts but I can add it to that PR instead if we don't go this route.


timwright12:
This seems good to me, can we merge @tlovett1?

PR #263, opened 2018-10-27
Add missing ref to "top"

leepeterson:
All section links reference an anchor which seems to be missing from the codebase: `#top`. The existing `nav` element with class `top-nav` appears to be the best candidate.


tlovett1:
Hmmm, somehow `#top` seems to work anyway even though it isn't in the DOM. Can't seem to figure out how that works...

nate-allen:
`#anythinghere` anchors will take you to the top of the page if there isn't a specific element with that ID.

Issue #185, opened 2017-03-16
Archiving Branches


ivankruchkoff:
Would like to start a discussion here as to whether this is relevant. Thoughts?

ericmann:
This pattern was added to the best practices documentation to massage away a few individual concerns about using version control that, frankly, no longer apply. If the team is properly trained on how VCS works (either Git or otherwise), "archived branches" is fundamentally unnecessary.

I say this as the original proponent of the practice. 👍  to removing the documentation as irrelevant.

stevegrunwell:
I'm all for removing it. Git tags are useful for versioned releases or major milestones (e.g. "rolled out the new theme", "launched this major feature", etc.) but archiving `fix/spelling` clutters the repository.

If not a complete removal, the archiving should be scoped to only those that would truly benefit from tagging.

ivankruchkoff:
There's some active discussion in the 10up slack, reasons where it can be useful:
1) Beanstalk takes a long time to load for a CR.
Proposed solution: delete unused branches.

2) An incomplete feature still exists, and we want to avoid an engineer accidentally deleting the branch.
Proposed solution: anyone with a local copy can still restore the branch, anyone who knows what to grok can still find the commit for the branch and restore it, or we can archive tag only those branches.

3) Audit purposes, I think this is very project specific and shouldn't influence the way all project repos operate.

I think the biggest benefit for tags are for your CI system to tag a release so that a rollback is easy to occur.

tlovett1:
@ericmann what were the original concerns? I never really understood this. Here are the branch scenarios that I see:

1. Engineer is working on branch. Branch stays.
2. Engineer started on a branch but decided the code change was unnecessary. Engineer deletes branch.
3. Engineer finished a branch and merged into staging/master. Engineer deletes branch. Anyone who wants to find the branch should be able to do so via the commit log since the branch was merged without fast forward.

Am I missing something?

ericmann:
The original concerns were relevant to a specific team working on a specific client project (which was, at the time, the only project actively using multiple feature branches at the same time). The practice of archive branches was designed specifically for that project and adopted more broadly as it was the "best solution we have right now."

[...]

PR #271, opened 2019-01-08
Add archiving branches as part of cleanup

adamsilverstein:
### Proposing
* Archive branches as part of cleanup flow when deleting them.
* add exact git commands to clarify what should happen

### Benefits
Keeping an archive of merged branch as tag captures makes restoring branches much easier should the need arise and simplifies some repository forensics, eg tracking down specific commits from a squashed merge.

### DIFF
--- a/_includes/markdown/Version-Control.md
+++ b/_includes/markdown/Version-Control.md
@@ -83,12 +83,21 @@ In the event that VIP makes a change to the repository, we'll capture the diff o
 * Merging the branch to ```staging```, using a non-fast-forward merge
 * Merging the branch back to ```master```, again using a non-fast-forward merge
 
-#### Archiving and Deleting Branches
+#### Deleting or Archiving and Deleting Branches
 
-This workflow will inevitably build up a large list of branches in the repository. To prevent a large number of unused branches living in the repository, we'll archive and delete them after feature development is complete.
+This workflow will inevitably build up a large list of branches in the repository. To prevent a large number of unused branches living in the repository, we'll delete or archive and delete them after feature development is complete.
+
+### Deleting branches
+When projects use non-ff merges to master, we can safely delte feature branches because all commits are preserved and can be located from the merge commit.
 
 * Move to another branch (doesn't matter which): eg. `git checkout master`
-* Archive the branch: `git tag archive/branch-name branch-name` 
+* Delete the branch (both on local and remote): `git branch -D branch-name; git push :branch-name`
+
+### Archiving and Deleting branches
+When projects use squash merges to create a more streamlined history in master, we should archive branches before deleting them to preserve the commit history. Branch tag archives also prove a useful history of branches, in case a specific branch is needed later.
+
+* Move to another branch (doesn't matter which): eg. `git checkout master`
+* Archive the branch: `git tag archive/branch-name branch-name`
 * Delete the branch (both on local and remote): `git branch -D branch-name; git push :branch-name`
 * Push the archive tag: `git push origin archive/branch-name`
 
--- a/_includes/markdown/Version-Control.md
+++ b/_includes/markdown/Version-Control.md
@@ -88,7 +88,7 @@ In the event that VIP makes a change to the repository, we'll capture the diff o
 This workflow will inevitably build up a large list of branches in the repository. To prevent a large number of unused branches living in the repository, we'll delete or archive and delete them after feature development is complete.
 
 ### Deleting branches
-When projects use non-ff merges to master, we can safely delte feature branches because all commits are preserved and can be located from the merge commit.
+When projects use non-ff merges to master, we can safely delete feature branches because all commits are preserved and can be located from the merge commit.
 
 * Move to another branch (doesn't matter which): eg. `git checkout master`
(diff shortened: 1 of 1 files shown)


nicholasio:
Not so long ago we decided that archiving branches did not make sense anymore: https://github.com/10up/Engineering-Best-Practices/issues/185. I believe the main takeaway was that, the need for archiving branches would not make sense for most of the projects and it is still possible to restore code and track down commits even without tags.

ericmann:
As @nicholasio points out, this effectively reverts the changes introduced by #188 (explanation in #185). Feels like a move backwards in the quality of the documentation.

> Keeping an archive of merged branch as tag captures makes restoring branches much easier should the need arise 

Using non-fast-forward merges provides the same benefit without cluttering the tag list for a project.

> simplifies some repository forensics, eg tracking down specific commits from a squashed merge

Unless I'm missing something, squashed merges are already discouraged in the workflows section: https://10up.github.io/Engineering-Best-Practices/version-control/#merges

ryanwelcher:
Whether or not archiving branches is needed is going to be situational. In general, 10up doesn't do it anymore but that doesn't mean that other teams using this document have the same view. 

Perhaps this becomes a best practice on how to archive effectively if the choice is made to do so rather than prescribing if they should or not.

I like the idea of adding git commands, it may not be clear how to get a branch back from either a tag or merge commit.

ericmann:
> In general, 10up doesn't do it anymore but that doesn't mean that other teams using this document have the same view.

Archiving was only a practice at 10up because of a specific project and team that needed it. I was the one who came up with the practice at the time and originally standardized it in the docs here. It only ever made sense then for a _super_ limited use case and.

That's why we _removed_ it in the aforementioned issues and commits.

> I like the idea of adding git commands, it may not be clear how to get a branch back from either a tag or merge commit.

General guidance on using Git to create a new branch based on a merge commit or a tag would be useful, but should definitely be in a different section.

adamsilverstein:
Thanks for the feedback @nicholasio and @ericmann - I missed the previous discussion - happy to withdraw or rework the proposal if the original thinking still stands.

I brought this up after _working on the several projects where squash merges were explicitly required_ (see: Gutenberg). Squashed commits are excellent for keeping the commit history in master clean, and tagging branches seems like a simple way to maintain the branch history at the same time. 

> Using non-fast-forward merges provides the same benefit without cluttering the tag list for a project.

Right, so maybe we can explicitly say this is for when not using fast-forward. I don't understand why "cluttering the tag list" is an issue.

ryanwelcher (approved):

adamsilverstein:
@ericmann @ryanwelcher @nicholasio -

Updated PR clarifying when deleting and arching might be appropriate, and why. What do you think?

ericmann:
> I don't understand why "cluttering the tag list" is an issue.

I use the command line, so I don't care about clutter. Large tag lists, however, have crashed GUI based utilities other engineers use, particularly on large projects. Tags should be used to, you know, _tag commits_ with some specific meaning. A reference to a release, to production or otherwise, for example.

They provide meaningful context and should continue to do so. Creating a tag because of a merge is an antipattern that I 100% regret ever having suggested.

> What do you think?

At this stage, it's entirely up to you and the team. I came up with "archived branches" as a workflow to appease non-technical stakeholders on a specific VIP project. We codified it at the time because there were similarly less-than-technical engineers on the team who didn't fully understand version control. We removed the practice from the docs later because it never made sense in the first place and removing it was a good idea.

I think I've said more than enough here and am officially backing out of this conversation.

adamsilverstein:
> Tags should be used to, you know, tag commits with some specific meaning.

Archiving a branch before deleting it does just that: it creates a commit with a specific meaning: the branch being deleted and its commit history

> I came up with "archived branches" as a workflow to appease non-technical stakeholders on a specific VIP project

I appreciate that you regret having proposed this as a best practice :) I still think it makes sense in certain cases (the two main projects I am working on both use a squashed commit merge strategy) as explained in my PR. I've also seen archiving strategies used on other projects in non-10up contexts.

Thanks for your feedback and the historical context.

nicholasio (commented):
I haven't worked on a squash and merge workflow yet, but after reading a bit more about the workflow I can see a few benefits and I can see tagged branches being valuable mainly to keep the commit history.

I'm ok with this since is imposing the tagging workflow only for squash and merge workflow.

But before we merge this, @adamsilverstein what do you think about adding some guidance on when to use `--no-ff` versus `squash and merge`? I don't think we mention the squash and merge workflow anywhere in our docs

ryanwelcher:
I like this approach @adamsilverstein. I think we need to be careful to provide guidance in a Best Practice doc and not be too prescriptive. If there is a need to archive branches, now we have a recommended way to handle it.

We're not saying that doing so is the best approach, we're just providing guidance based on our experiences doing so.

adamsilverstein:
> But before we merge this, @adamsilverstein what do you think about adding some guidance on when to use --no-ff versus squash and merge? I don't think we mention the squash and merge workflow anywhere in our docs

That seems useful, summarizing pros/cons of each approach; thanks for the suggestion. 

I am going to start by opening an issue to add that information so I can get this merged and solicit wider feedback on squash vs. -no-ff

Issue #274, opened 2019-01-11
Fix Back to Top


circlecube:
Currently the back to top links in each section do not take the user to the top of the page.

circlecube:
This looks like it is related to the sticky header. The nav with #top id is in the header, so clicking the link doesn't take you to the top because it's already on the page. We'll need to add an anchor to the top of the page after the header (likely add an id to an existing element)

PR #275, opened 2019-01-11
move #top to template header instead of page header 

circlecube:
(since page header is sticky)

Fixes issue #274


colorful-tones:
Just tested this and working nicely. 👍

Issue #300, opened 2019-03-21
Broken links for Design section


jeffpaul:
The `Design` link in the header links to https://10up.github.io/Engineering-Best-Practices/structure/ (this resolves correctly) while the `Project Design` link in the Table of Contents links to https://10up.github.io/Engineering-Best-Practices/design/#top (this 404s).  I would normally open a PR to fix the `Project Design` link, but from a URL naming structure it looks like both should use the https://10up.github.io/Engineering-Best-Practices/design/ and update the docs foldering to use /design/ instead of /structure.  I'll leave it up to discussion here on the best approach to fix this.

TheLastCicada:
This is still very broken and seems like an easy fix.....

PR #302, opened 2019-03-30
fix: Update broken intro page design menu URL's

TylerBarnes:
Thought it would be good to write a couple lines of code on my first week ;)

It appears the design page was renamed to structure at some point. The top menu looks like it generates links dynamically but these are hardcoded. The #privacy section also seems to not exist so part of updating these was removing that link.

Edit: You can find the broken links under the "Project Design" section on the [home page](https://10up.github.io/Engineering-Best-Practices/).

### DIFF
--- a/index.md
+++ b/index.md
@@ -105,13 +105,12 @@ updated: 6 Oct 2014
 	</div>
 
 	<div class="col">
-		<h3><a href="{{ site.baseurl }}/design/#top">Project Design</a></h3>
+		<h3><a href="{{ site.baseurl }}/structure/#top">Project Design</a></h3>
 		<ul>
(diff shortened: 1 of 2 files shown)


jeffpaul:
Resolves #300

TylerBarnes:
Oops, didn't realize there was already an open PR for this!

TylerBarnes:
@tylercherpak this has been broken since I started at 10up. Just got pinged about it from @jeffpaul. How do we get this merged? More info at #300

helen:
It's because I made #262 for #229 but the latter still isn't in, let's just merge this and fix what exists and worry about the file name later (and add the TOC item to #229 and finally finish it...)

TylerBarnes:
Thanks @helen !

PR #304, opened 2019-05-14
Adds Debounce and Throttle to JS Performance section

ryansommers:
### Description of the Change

Adds Debounce and Throttle descriptions and examples to JavaScript Performance section.

### DIFF
--- a/_includes/markdown/JavaScript.md
+++ b/_includes/markdown/JavaScript.md
@@ -278,7 +278,7 @@ hideButton.addEventListener( 'click', () => {
 
 Notice how, in cached versions, we are pulling the menu selection out of the event listener so it only happens once. The cached version is, not surprisingly, the [fastest way to handle this situation](https://jsperf.com/dom-selection-caching).
 
-#### Event Delegation
+### Event Delegation
 
 Event delegation is the act of adding one event listener to a parent node to listen for events bubbling up from its children. This is much more performant than adding one event listener for each child element. Here is an example:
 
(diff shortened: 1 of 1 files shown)


jonathantneal:
Nice. This is handy information. Well written, @ryansommers!

_includes/markdown/JavaScript.md
```diff
@@ -304,6 +304,77 @@ document.getElementById( 'menu' ).addEventListener( 'click', ( e ) => {
...
+
+Browser events such as scrolling, resizing, and cursor movements happen as fast as possible and can cause performance issues. Debouncing and throttling our functions helps us increase performance by controlling the rate at which an event listener calls them. 
+
+Debouncing a function will prevent it from being called again until a defined amount of time has passed, i.e., execute this function if 200ms has passed since it was last called. A common use case would be when resizing a browser window; we can apply classes or move elements after the resize has happened.
+
+Throttling a function will cause it to only be called a maximum number of times over a defined period of time, i.e., only execute this function once every 50ms. A common use case would be when scrolling a browser window; we may want visual effects as a scroll is happening.
+
+Here are example debounce and throttle functions, though note that some of our recommended utility libraries already have similar functions, such as Underscore's _.debounce() and _.throttle(). 
+
+```javascript
```
timwright12: Can you move this under the debounce paragraph so we don't have two code blocks next to eachother? Then add another note for something like, "this is a debounce example". Also it would be nice to get some comments in the code talking about what's actually going on here

_includes/markdown/JavaScript.md
```diff
@@ -304,6 +304,77 @@ document.getElementById( 'menu' ).addEventListener( 'click', ( e ) => {
...
+		}
+	};
+};
+
+const someDebouncedFunction = () => console.info( 'Here’s a debounced function.' );
+
+window.addEventListener( 'resize', debounce( someDebouncedFunction, 200 ) );
+```
+
+```javascript
```
timwright12: Can we add some comments into this code to describe what's going on?

timwright12 (changes requested):

_includes/markdown/JavaScript.md
```diff
@@ -304,6 +304,77 @@ document.getElementById( 'menu' ).addEventListener( 'click', ( e ) => {
 
 You may be wondering why we don't just add one listener to the `<body>` for all our events. Well, we want the event to *bubble up the DOM as little as possible* for [performance reasons](https://jsperf.com/event-delegation-distance). This would also be pretty messy code to write.
 
+### Debounce and Throttle
+
+Browser events such as scrolling, resizing, and cursor movements happen as fast as possible and can cause performance issues. Debouncing and throttling our functions helps us increase performance by controlling the rate at which an event listener calls them. 
+
+Debouncing a function will prevent it from being called again until a defined amount of time has passed, i.e., execute this function if 200ms has passed since it was last called. A common use case would be when resizing a browser window; we can apply classes or move elements after the resize has happened.
```
timwright12, line 312: Let's link to a debouncing documentation resource

_includes/markdown/JavaScript.md
```diff
@@ -304,6 +304,77 @@ document.getElementById( 'menu' ).addEventListener( 'click', ( e ) => {
...
+
+Throttling a function will cause it to only be called a maximum number of times over a defined period of time, i.e., only execute this function once every 50ms. A common use case would be when scrolling a browser window; we may want visual effects as a scroll is happening.
```
timwright12: Let's link to a trottling documentation resource

_includes/markdown/JavaScript.md
ryansommers: Linked to underscore, lodash, and CSS Tricks explanations at the end. Let me know if you were thinking something else.

_includes/markdown/JavaScript.md
ryansommers: Linked to underscore, lodash, and CSS Tricks explanations at the end. Let me know if you were thinking something else.

timwright12:
@ryansommers seems like `requestAnimationFrame()` is now the preferred solution for this kind of thing to get the listeners executing on paint instead of a random number interval. Can you look into that and update accordingly? https://www.afasterweb.com/2017/09/26/performance-basics-throttling/

If there are instances where we still want a traditional throttle/debounce we can note those as well.

ryansommers:
@timwright12 it looks like `requestAnimationFrame()` is specific to animations and would only be appropriate as a throttle function, not debounce. And even with throttle, it's shooting for throttling to 60fps, which isn't necessarily what we'd always want anyway. There's a good pros/cons list at https://css-tricks.com/debouncing-throttling-explained-examples/#article-header-id-9

If we want to include `requestAnimationFrame()` I'd think it should another separate function or example for animation specific performance rather than a wholesale replacement of `throttle()`

timwright12:
@ryansommers seems like a pretty good distinction in use cases between those. How about this: let's lose the large code blocks and update the content to three sections, 1. debounce, 2. throttle, and 3.requestAnimationFrame - inside each we can have a description of the method, information about when to use them, and links to some example code. I know we're starting to see requestAnimationFrame replacing debounce/throttle, so if we can clear that up it would be great.

After that, let's update it a little and post something on the internal blog so more people see the update. If the code snippets are all utility enough we can start to spin up a new component for utilities - if we're better off using something like lodash with something custom for requestAnimationFrame we can note that in a blog post with some code samples

ryansommers:
@timwright12 pushed updates and simplified, ready for another review.

timwright12:
Looks good to me!

PR #311, opened 2019-09-05
Feature/coding style updates 2019 09

mmcachran:
### Checklist:

- [x] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have added tests to cover my change.
- [x] All new and existing tests passed.

### Applicable Issues

None

### Changelog Entry

Added coding style documentation for easier debugging and code cleanup.

### DIFF
--- a/_includes/markdown/PHP.md
+++ b/_includes/markdown/PHP.md
@@ -976,7 +976,7 @@ function is_valid_api_response( array $data = [] ) {
 }
 ```
 
-This function checks to ensure all expected data was returned from the API. Now imagine there’s a lot of data you have to check for. Adding more isset checks will fix this issue, but makes debugging hard since you have to check each index individually. This can be refactored to make debugging easier.
+This function checks to ensure all expected data was returned from the API. Consider there’s a lot of data you have to check for. Adding more isset checks will fix this issue, but makes debugging hard since you have to check each index individually. This can be refactored to make debugging easier.
 
 ```php
 /**
--- a/_includes/markdown/PHP.md
+++ b/_includes/markdown/PHP.md
@@ -919,7 +919,7 @@ Sessions should never be stored in the database. This introduces extra data into
 
 ### Avoid nested logic
 
-Nesting logic makes code hard to read. The following is an example of the [wp_unslash](https://codex.wordpress.org/Function_Reference/wp_unslash).
+Nesting logic makes code hard to read. The following is an example of the [wp_unslash](https://codex.wordpress.org/Function_Reference/wp_unslash) function in WP Core.
 
 ```php
 function wp_slash( $value ) {
--- a/_includes/markdown/PHP.md
+++ b/_includes/markdown/PHP.md
@@ -988,10 +988,10 @@ This function checks to ensure all expected data was returned from the API. Cons
 function is_valid_api_response( array $data = [] ) {
 	// Holds data we must have in the API response for it to be considered valid.
 	$required_indices = [
-		‘required_index’,
-‘another_required_index’,
-‘yet_another_required_index’,
-‘and_yet_another_required_index’,
+		'required_index',
+'another_required_index',
+'yet_another_required_index',
+'and_yet_another_required_index',
 ];
 
 // Loop through required indices to check each in the API response.
--- a/_includes/markdown/PHP.md
+++ b/_includes/markdown/PHP.md
@@ -916,3 +916,98 @@ Instead of sessions, use cookies or client-side storage APIs if possible. In add
 If sessions must be used, create them conservatively. Don't create sessions for every visitor. Limit sessions to the smallest group that needs them: logged-in editors and admins, or visitors using a particular feature.
 
 Sessions should never be stored in the database. This introduces extra data into a storage system that's not meant for that volume. Database session libraries also rely on PHP code which can't match the performance of PHP's native session handlers. PHP extensions for Memcache and Redis allow sessions to be stored in these in-memory datastores and are a good solution for sessions when multiple webservers are present.
+
+### Avoid nested logic
+
+Nesting logic makes code hard to read. The following is an example of the [wp_unslash](https://codex.wordpress.org/Function_Reference/wp_unslash).
+
+```php
+function wp_slash( $value ) {
+    if ( is_array( $value ) ) {
+        foreach ( $value as $k => $v ) {
+            if ( is_array( $v ) ) {
+                $value[ $k ] = wp_slash( $v );
+            } else {
+                $value[ $k ] = addslashes( $v );
+            }
+        }
+    } else {
+        $value = addslashes( $value );
+    }
+ 
+    return $value;
+}
+```
+
+This function could be rewritten to make it easier to understand and remove nested logic.
+
+```php
+function wp_slash( $value ) {
+  // Bail early if value is not an array.
+  if ( ! is_array( $value ) ) {
+    return addslashes( $value );
+  }
+  
+  // Loop through values and add slashes to each value.
+  foreach ( $value as $k => $v ) {
+    // Sanitize value or values.
+   $values[ $k ] = is_array( $v ) ? wp_slash( $v ) : addslashes( $v );
+  }
+
+  return $values;
+}
+```
+
+With the refactoring from above, the nested logic is removed and "bail early" logic is being used to return if the value is not an array. A ternary is then used to determine if the function needs to be recursively called and, finally, the values are returned.
+
+### Sanity checking data
+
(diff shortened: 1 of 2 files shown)


_includes/markdown/PHP.md
```diff
@@ -916,3 +916,98 @@ Instead of sessions, use cookies or client-side storage APIs if possible. In add
...
+ * @param array $data The data to check.
+ * @return bool              True if all required pieces of data exist, false otherwise.
+ */
+function is_valid_api_response( array $data = [] ) {
+	// Holds data we must have in the API response for it to be considered valid.
+	$required_indices = [
+		‘required_index’,
+‘another_required_index’,
+‘yet_another_required_index’,
+‘and_yet_another_required_index’,
```
aaemnnosttv: Curly quotes!

mmcachran: 😱 Fixed in the newest commit.

jeffpaul:
@aaemnnosttv any updated review feedback based on @mmcachran latest changes here?

_includes/markdown/PHP.md
```diff
@@ -916,3 +916,98 @@ Instead of sessions, use cookies or client-side storage APIs if possible. In add
...
+
+```php
+function wp_slash( $value ) {
+  // Bail early if value is not an array.
+  if ( ! is_array( $value ) ) {
+    return addslashes( $value );
+  }
+  
+  // Loop through values and add slashes to each value.
+  foreach ( $value as $k => $v ) {
```
aaemnnosttv: The first block could be completely avoided by casting `$value` to an array in the `foreach`, further simplifying the code. It also adds support for passing a simple object as the value.

_includes/markdown/PHP.md
```diff
@@ -916,3 +916,98 @@ Instead of sessions, use cookies or client-side storage APIs if possible. In add
...
+
+```php
+/**
+ * Checks an API response to ensure all required pieces of data was returned.
+ *
+ * @param array $data The data to check.
+ * @return bool              True if all required pieces of data exist, false otherwise.
+ */
+function is_valid_api_response( array $data = [] ) {
+	return isset( $data['required_index'], $data['another_required_index'], $data['yet_another_required_index'], $data['and_yet_another_required_index']  );
```
aaemnnosttv:
```suggestion
	return isset(
		$data['required_index'],
		$data['another_required_index'],
		$data['yet_another_required_index'],
		$data['and_yet_another_required_index'] 
	);
```

_includes/markdown/PHP.md
```diff
@@ -916,3 +916,98 @@ Instead of sessions, use cookies or client-side storage APIs if possible. In add
...
+ * Checks an API response to ensure all required pieces of data was returned.
+ *
+ * @param array $data The data to check.
+ * @return bool              True if all required pieces of data exist, false otherwise.
+ */
+function is_valid_api_response( array $data = [] ) {
+	// Holds data we must have in the API response for it to be considered valid.
+	$required_indices = [
+		'required_index',
+'another_required_index',
```
aaemnnosttv:
```suggestion
		'another_required_index',
```

_includes/markdown/PHP.md
```diff
@@ -916,3 +916,98 @@ Instead of sessions, use cookies or client-side storage APIs if possible. In add
...
+'yet_another_required_index',
```
aaemnnosttv:
```suggestion
		'yet_another_required_index',
```

_includes/markdown/PHP.md
```diff
@@ -916,3 +916,98 @@ Instead of sessions, use cookies or client-side storage APIs if possible. In add
...
+'and_yet_another_required_index',
```
aaemnnosttv:
```suggestion
		'and_yet_another_required_index',
```

aaemnnosttv (commented):
Left a comment that I think could simplify things a bit more and introduce another strategy to use for reducing nesting in addition to conditionals in general, by standardizing/limiting the number of types you need to be concerned with.

A few other suggestions related to formatting.

jeffpaul:
@mmcachran back to you on this one

_includes/markdown/PHP.md
mmcachran: Hi, @aaemnnosttv - Wouldn't casting the value to an array break the response because instead of a string the return would be an array?

aaemnnosttv: I wasn't suggesting the _return value_ be cast, instead that the first conditional block is unnecessary if we coerce the `$value` parameter to always be an array 😉

mmcachran: right - I mean if we skip the first block the use this as the parameter
$array = [
  'test',
]

it would return this

$array = [
  0 => [
    'test',
  ]
];

since we'd be coercing the value to be an array it would be returned as an array :)

A simpler example of how this would break the functionality would be passing in just a string like $s = 'test'; would return as $s = [ 0 => 'test' ];

You are right that the function could be updated to support objects, but the example was just to show how rethinking the function logic could improve readability.

aaemnnosttv: You're right, I didn't realize it returns the same type its given - the first block is necessary there.

One thing I did catch is that `$values` needs to be initialized or handled in the return at the end to avoid an undefined variable notice if the function is called with an empty array.

_includes/markdown/PHP.md
```diff
@@ -916,3 +916,103 @@ Instead of sessions, use cookies or client-side storage APIs if possible. In add
...
+    return addslashes( $value );
+  }
+  
+  // Loop through values and add slashes to each value.
+  foreach ( $value as $k => $v ) {
+    // Sanitize value or values.
+    $values[ $k ] = is_array( $v ) ? wp_slash( $v ) : addslashes( $v );
+  }
+
+  return $values;
```
aaemnnosttv: If `$values` is null (uninitialized), return an empty array.

```suggestion
  return $values ?? [];
```

_includes/markdown/PHP.md
```diff
@@ -916,3 +916,103 @@ Instead of sessions, use cookies or client-side storage APIs if possible. In add
...
+	if ( isset( $data[ $index ] ) ) {
+		continue;
+	}
+	
+	// Bail early because a required piece of data wasn’t returned.
+	return false;
+}
+
+// All required pieces of data exist.
+return true;
```
aaemnnosttv: Missing 1 indent here

aaemnnosttv (changes requested):
Thanks @mmcachran !

Just one formatting correction to make and I added a suggestion for handling an edge case in your `wp_slash` function.

_includes/markdown/PHP.md
mmcachran: :derp: let me fix that real quick

_includes/markdown/PHP.md
mmcachran: This should be taken care of by initializing $values before it is used.

mmcachran:
Thanks, @aaemnnosttv! Back to you for review 🙂

bengreeley:
@aaemnnosttv @mmcachran I'm good with this addition to the best practices to avoid nested logic/if statements.

@tlovett1 thoughts?

bengreeley (approved):

aaemnnosttv (approved):
Nothing else from me 😄 nice work @mmcachran 👍

tlovett1:
Hey @mmcachran thanks for putting this together. The first part about nested logic definitely makes sense to me. The second part about checking array keys feels a bit subjective. Can you split that off into a separate PR?

_includes/markdown/PHP.md
```diff
@@ -916,3 +916,105 @@ Instead of sessions, use cookies or client-side storage APIs if possible. In add
...
+function wp_slash( $value ) {
+    // Bail early if value is not an array.
+    if ( ! is_array( $value ) ) {
+        return addslashes( $value );
+    }
+
+    $values = [];
+
+    // Loop through values and add slashes to each value.
+    foreach ( $value as $k => $v ) {
```
christianc1: To follow our other best practices, can you expand this variable names to `$key` and `$value` instead of `$k` and `$v`.  You'll also need to change the initial argument to something like `$input`.

christianc1 (changes requested):

_includes/markdown/PHP.md
mmcachran: _updated_

christianc1 (approved):

jeffpaul:
@tlovett1 this looks like it may be good to merge, yeah?

jeffpaul:
@tlovett1 any concerns with us merging this in?

PR #322, opened 2019-12-09
Add section for Memcached and Redis

TheLastCicada:
New section in Systems.md that discusses the hosting considerations for Memcached and Redis.  
This section is fully written and is ready for review.


_includes/markdown/Systems.md
```diff
@@ -252,3 +253,94 @@ Once tuning moves beyond these main items, performance improvements will be mino
...
+
+Starting in memcached version 1.4.2, the item size is configurable via the `-I` option.  Setting the item size higher than 1 MB is not recommended unless necessary as memcached becomes less efficient as the item size increases, meaning it will take more memory to store the same amount of data.  If memory is available to accommodate a larger cache size, it is worth considering raising the item size to prevent the alloptions array from ever becoming a problem.  
+
+#### Connecting WordPress to Memcached
+
+WordPress connects to memcached through an object-cache.php drop-in plugin file placed in the wp-content folder.  The object-cache.php file will leverage a php extension to handle communicating with memcached.  There are two PHP extensions commonly used, confusingly named [php-memcached](https://pecl.php.net/package/memcache) and [php-memcached](https://pecl.php.net/package/memcached).  It is important to match the object-cache.php file with the right PHP extension, and, while there’s many object cache files that can work, 10up mostly uses one of the following:
```
nicholasio: in `confusingly named [php-memcached](https://pecl.php.net/package/memcache) and [php-memcached](https://pecl.php.net/package/memcached).` The first one is php-memcache instead of php-memcached.

_includes/markdown/Systems.md
```diff
@@ -252,3 +253,94 @@ Once tuning moves beyond these main items, performance improvements will be mino
...
+
+#### Item Size
+
+By default, memcached accepts items (key + value) of 1 MB or less. In most scenarios, this is fine and plenty of space. However, on larger WordPress sites or sites with a number of plugins, the “alloptions” array combining all autoloaded rows from the wp_options table can exceed 1 MB.  This array will be stored in memcached if memcached is in use for the object cache, unless it exceeds the memcached item size limit.  When this array is larger than the memcached item size, it can cause all sorts of odd issues and inconsistent performance.  
+
+Starting in memcached version 1.4.2, the item size is configurable via the `-I` option.  Setting the item size higher than 1 MB is not recommended unless necessary as memcached becomes less efficient as the item size increases, meaning it will take more memory to store the same amount of data.  If memory is available to accommodate a larger cache size, it is worth considering raising the item size to prevent the alloptions array from ever becoming a problem.  
+
+#### Connecting WordPress to Memcached
```
nicholasio, line 297: I've had trouble in the past choosing the right one for WP Local Docker and I've seen people with the same problem. It might make sense to call out which one is the correct one to use with WP Local Docker.

_includes/markdown/Systems.md
```diff
@@ -252,3 +253,94 @@ Once tuning moves beyond these main items, performance improvements will be mino
...
+
+Memcached and Redis are in memory data stores that are used for the WordPress object cache. Implementation of the object cache in WordPress code is covered extensively in the [PHP Performance](https://10up.github.io/Engineering-Best-Practices/php/#performance) section of the Best Practices and this section will focus on the hosting and setup of memcached and Redis.
+
+Memcached and Redis are used by WordPress as simple, in memory key-value stores. By being in memory and not having the possibility for complex queries, these tools provide blazing fast retrieval of data, usually in less than 1 millisecond. Common use cases of these data stores are:
+
+* Caching results of an external API call
+* Caching the results of a complex query
+* Storing full or partial pages 
+
+Items stored in these caching technologies are made up of 3 things: a key, a value, and an optional expiration time.  After the expiration time, items are evicted from the cache.  Items without an expiration time will persist in the cache until the cache is full, at which time the cache will evict the least recently used (LRU) item each time a new item needs to be stored.
```
tott: Possibly a good idea to mention that while LRU is commonly the default there are other eviction policies

_includes/markdown/Systems.md
```diff
@@ -252,3 +253,94 @@ Once tuning moves beyond these main items, performance improvements will be mino
...
+
+A common misconception is that a full cache pool is a problem and should be avoided, or that a full cache means a bigger cache pool is needed. In reality, a full cache pool is the normal state of the cache with WordPress and means nothing.  No matter how big the cache pool, it will eventually become full.  This is because there will always be some cached items without an expiration time that will persist in the cache forever, until the cache is full and the LRU policy evicts it to make room for a new item.
+
+Even a very active cache will not stress a modern CPU, so a caching server can perform very well even with a single CPU core.  Similarly, since the cache activity is all in memory (unless persistence is enabled with Redis), a small and slow hard drive will work fine.  As long as network bandwidth is fast, latency is low, and enough RAM is available for the cache pool, nearly any server can host the cache. 
+
```
tott, line 276: Should we give some pointers on how to determine issues. ie. memcache stats phpmemcacheadmin/phpredisadmin and looking at ratios between set/get/evictions in context?

_includes/markdown/Systems.md
```diff
@@ -252,3 +253,94 @@ Once tuning moves beyond these main items, performance improvements will be mino
...
+#### Item Size
+
+By default, memcached accepts items (key + value) of 1 MB or less. In most scenarios, this is fine and plenty of space. However, on larger WordPress sites or sites with a number of plugins, the “alloptions” array combining all autoloaded rows from the wp_options table can exceed 1 MB.  This array will be stored in memcached if memcached is in use for the object cache, unless it exceeds the memcached item size limit.  When this array is larger than the memcached item size, it can cause all sorts of odd issues and inconsistent performance.
```
tott: This is only one example. We should point out that this might also apply for other scenarios like caching remote calls, query results or html fragments.

_includes/markdown/Systems.md
tott: maybe just link to tools section below

_includes/markdown/Systems.md
TheLastCicada: I can't tell you how many times I looked at it to make sure I typed it right and I still screwed it up.  Change coming in a moment.

_includes/markdown/Systems.md
TheLastCicada: I have no idea which is the right one to use with WP Local Docker and really it should be compatible with both.  I think the documentation for wp local docker would be the better place to document this.

_includes/markdown/Systems.md
TheLastCicada: Good call

_includes/markdown/Systems.md
TheLastCicada: I decided not to wade into that in this round as it is so different from site to site - there's not good numbers or direction I could come up with that was universally applicable.  A membership site will have a much different hit rate than a blog site.  A smaller cache pool will have a higher eviction rate than a larger pool, but the hit rate might be fine.  A large eviction rate doesn't mean much if you don't set expiration times, which is also fine.  I think I'm going to leave it as is - the tools section is fairly close to this one and I feel like all the sections written are important and connected.

_includes/markdown/Systems.md
TheLastCicada: I'll address this in the next commit.

jasondewitt (approved):
looks good to me, save for the stuff we talked about on slack.

PR #323, opened 2019-12-14
New Load Balancing section in Systems

TheLastCicada:
### Description of the Change

New section in Systems about Load Balancing and multi-server architectures.  

### Benefits

Common questions about WordPress at scale are about how to load balance and what to be concerned about. This covers load balancing and going from a single server to multiple servers.  

### Possible Drawbacks

Could go further into cloud architecture, but limits scope to what is related to load balancing.


jeffpaul:
@jasondewitt @tott any plans to review this to potentially add this new section?

PR #331, opened 2020-04-08
Updating code style & React Best Practices

nicholasio:
### Description of the Change

- Updates eslint to use the new eslint config and also updates the js snippets to match the new rules. Additionally, it sets up eslint to be able to lint and format code within the snippets.
- Several updates to the React best practices.


_includes/markdown/JavaScript.md
```diff
@@ -1,3 +1,9 @@
+<h2 id="code-style" class="anchor-heading">Code Style, Tooling & Documentation {% include Util/link_anchor anchor="code-style" %} {% include Util/top %}</h2>
```
nicholasio, line 1: Most of the changes on this file are automatic fixes by the linter.

_includes/markdown/JavaScript.md
```diff
@@ -158,38 +164,36 @@ export const example = 66;
...
 
 When a script is not wrapped in a closure, the current context or ```this``` is actually ```window```:
 
 ```javascript
-console.log( this === window ); // true
+console.log(this === window); // true
 
-for ( var i = 0; i < 9; i++ ) {
-    // Do stuff
+for (var i = 0; i < 9; i += 1) {
```
bengreeley: @nicholasio Is there a reason why we'd change i++ to i+=1 in a for loop in Javascript? The ++ syntax is very common in programming for loops, so just curious if there's a reasoning here.

timwright12: I haven't researched this in a long time, but I think i += 1 is a slightly faster iteration method

nicholasio: yes, it's mostly an eslint thing, https://eslint.org/docs/rules/no-plusplus and https://github.com/airbnb/javascript#variables--unary-increment-decrement

timwright12: Also, there's a var here we should update

_includes/markdown/JavaScript.md
```diff
@@ -368,21 +370,15 @@ An essential part of a GraphQL API is an API schema. GraphQL requires a human-re
 If you are choosing to use GraphQL on a WordPress project, it is recommended to use the WPGraphQL plugin. This plugin will return WordPress data in JSON format through a GraphQL endpoint - in many cases you won’t need to write the schema yourself. This will give you all the benefits of concatenating your data requests as well as easy access to your data as it is output by WordPress. You can retrieve your data by passing a query directly into a simple fetch request:
 
 ```javascript
-fetch( '/graphql', {
-    method: 'POST',
-    headers: { 'Content-Type': 'application/json' },
-    body: JSON.stringify({ query: "[your query string goes here]" }),
-} ).then( res => res.json() );
+fetch('/graphql', {
```
timwright12, line 448: Do we still need this GraphQL stuff? We really don't use it.

_includes/markdown/JavaScript.md
bengreeley: While it's minor in the grand scheme of things, it seems a little odd to me (likely because I'm not a Javascript engineer ;-) ). I'm not sure if I'm the only engineer that would think this, but I'd like to suggest we add the following option to our ESLint, otherwise I see this as something that a programmer would use naturally since it's so common and need to revise their code for minor benefit: `"allowForLoopAfterthoughts": true`

joesnellpdx: I agree with @bengreeley . This will likely throw people off. I'd even recommend providing a 'note' below this block or a comment with in it for those that see this and are wondering the same question. 

OR - not using a loop to define your closure and avoid confusion in this instance.

joesnellpdx (commented):
@nicholasio 

Looks good to me. I left one comment on the closure / loop discussion.

The only thing I'd consider adding is best practices in use of the module pattern (or guidance on how add custom js files to a standard project). 

I see classes being used when a simple module pattern (namespaces or data structures) seems like a preferred option.

See: https://github.com/getify/You-Dont-Know-JS/blob/f6e1e05bc2fb212bc38acfeb7461804d48975728/scope-closures/ch8.md

Example: 
```
export default function myFunction(param1, param2) {
  // code here
}
``` 

Marking as approved as I see nothing blocking.

joesnellpdx (approved):

_includes/markdown/JavaScript.md
```diff
@@ -283,23 +287,21 @@ Notice how, in cached versions, we are pulling the menu selection out of the eve
...
-          // Do stuff with a parent.
-          target = target.parentNode;
-        }
-      }
-    }
-
-} );
+document.getElementById('menu').addEventListener('click', (e) => {
+	const { currentTarget } = e;
+	let { target } = event;
```
aaemnnosttv: Shouldn't these both be `e` or `event`? Seems like `event` is currently undefined.

_includes/markdown/JavaScript.md
nicholasio: Make sense! I create this issue to track whatelse we might want to week in our eslint config: https://github.com/10up/eslint-config/issues/20

I'll disable this here for now but will get this updated in eslint config at some point.

_includes/markdown/JavaScript.md
nicholasio: We do use GraphQL on some projects so I think it's work having something about it and we do have plans on using it more!.

I can review/update the graphql section on another PR if needed.

_includes/markdown/JavaScript.md
nicholasio: Good catch! I'll update this.

_includes/markdown/JavaScript.md
nicholasio: Done.

nicholasio:
> @nicholasio
> 
> Looks good to me. I left one comment on the closure / loop discussion.
> 
> The only thing I'd consider adding is best practices in use of the module pattern (or guidance on how add custom js files to a standard project).
> 
> I see classes being used when a simple module pattern (namespaces or data structures) seems like a preferred option.
> 
> See: https://github.com/getify/You-Dont-Know-JS/blob/f6e1e05bc2fb212bc38acfeb7461804d48975728/scope-closures/ch8.md
> 
> Example:
> 
> ```
> export default function myFunction(param1, param2) {
>   // code here
> }
> ```
> 
> Marking as approved as I see nothing blocking.

Added a section for modules, let me know if it is along the lines of what you were thinking. cc @timwright12 that we would probably want to update some of the fe components to be more inlined with this new section.

joesnellpdx:
> Added a section for modules, let me know if it is along the lines of what you were thinking.

@nicholasio This is GREAT! I suspect it will spark some conversation and may be enhanced by an internal post about it...  but all for the better / general knowledge bump for the team!

Thanks for adding!

jeffpaul:
Looks like all requested changes have been made.  Anything blocking this PR from being merged at this point?

PR #356, opened 2021-02-22
add paragraph about meta_queries and NOT_EXISTS

### LINKED ISSUE #355: Add paragraph on the performance of meta_queries using EXISTS/NOT_EXISTS
**Is your enhancement related to a problem? Please describe.**

It is very common that developers introduce new `meta_keys` without setting default values for posts that already existed in the database, or that they delete `meta_keys` when the value is not needed instead of setting it to a default or false value.

This becomes an issue on larger database tables when users need to find some information by filtering not only by the meta_key and value but also sometimes by checking if that `meta_key` _does not exist_ at all. This use case is covered in `WP_Meta_Query`, allowing for the use of `compare => 'NOT EXISTS'`

This solution should, however, be **discouraged**, as it generates very expensive queries that can really hammer a database performance.

Example:
- Database with 253k rows in wp_posts and 5 million rows in wp_postmeta
- There are 51,406 rows with the `_publicize_pending` meta_key in wp_postmeta

```
/**
 * Using NOT_EXISTS in a Meta Query generates a LEFT JOIN clause
 * with an IS NULL check in WHERE
 *
 * Rows searched: 268,358 (all rows in the join)
 */
$query = new \WP_Query(
	array(
		'suppress_filters' => true,
		'meta_query'       => array(
			array(
				'key'     => '_publicize_pending',
				'compare' => 'NOT EXISTS',
			),
		),
	) );
```
Generates:
```
MySQL [wordpress]> EXPLAIN SELECT SQL_CALC_FOUND_ROWS  wp_posts.ID FROM wp_posts  
    -> LEFT JOIN wp_postmeta ON (wp_posts.ID = wp_postmeta.post_id AND wp_postmeta.meta_key = 'some_key' ) 
    -> WHERE 1=1  
    -> AND ( wp_postmeta.post_id IS NULL ) 
    -> GROUP BY wp_posts.ID ORDER BY wp_posts.post_date DESC LIMIT 0, 10;
+----+-------------+-------------+------------+-------+-----------------------------------------------------------------+---------+---------+-----------------------+--------+----------+---------------------------------+
| id | select_type | table       | partitions | type  | possible_keys                                                   | key     | key_len | ref                   | rows   | filtered | Extra                           |
+----+-------------+-------------+------------+-------+-----------------------------------------------------------------+---------+---------+-----------------------+--------+----------+---------------------------------+
|  1 | SIMPLE      | wp_posts    | NULL       | index | PRIMARY,post_name,type_status_date,post_parent,post_author,guid | PRIMARY | 8       | NULL                  | 268339 |   100.00 | Using temporary; Using filesort |
|  1 | SIMPLE      | wp_postmeta | NULL       | ref   | post_id,meta_key,meta_key_meta_value                            | post_id | 8       | wordpress.wp_posts.ID |     19 |   100.00 | Using where; Not exists         |
+----+-------------+-------------+------------+-------+-----------------------------------------------------------------+---------+---------+-----------------------+--------+----------+---------------------------------+
```

In contrast, this query checking for a specific value in the declared `meta_key` only needs to read 89,970 rows, or roughly 70% fewer rows.

```
/**
 * Querying only for the meta_key and value generates an INNER JOIN clause
 * with a key/value match in WHERE
 *
 * Rows searched: 89,970
 */
$query = new \WP_Query(
	array(
		'suppress_filters' => true,
		'meta_query'       => array(
			array(
				'key'     => '_publicize_pending',
				'value'     => '1',
				'compare' => '!=',
			),
		),
	) );
```
Generates:
```
MySQL [wordpress]> EXPLAIN SELECT wp_posts.ID
    -> FROM wp_posts
    -> INNER JOIN wp_postmeta
    -> ON ( wp_posts.ID = wp_postmeta.post_id )
    -> WHERE 1=1
    -> AND ( ( wp_postmeta.meta_key = '_publicize_pending'
    -> AND wp_postmeta.meta_value != '1' ) );
+----+-------------+-------------+------------+--------+--------------------------------------+---------------------+---------+-------------------------------+-------+----------+-------------+
| id | select_type | table       | partitions | type   | possible_keys                        | key                 | key_len | ref                           | rows  | filtered | Extra       |
+----+-------------+-------------+------------+--------+--------------------------------------+---------------------+---------+-------------------------------+-------+----------+-------------+
|  1 | SIMPLE      | wp_postmeta | NULL       | range  | post_id,meta_key,meta_key_meta_value | meta_key_meta_value | 1534    | NULL                          | 89970 |   100.00 | Using where |
|  1 | SIMPLE      | wp_posts    | NULL       | eq_ref | PRIMARY                              | PRIMARY             | 8       | wordpress.wp_postmeta.post_id |     1 |   100.00 | Using index |
+----+-------------+-------------+------------+--------+--------------------------------------+---------------------+---------+-------------------------------+-------+----------+-------------+
2 rows in set, 1 warning (0.001 sec)
```

moraleida:
### Description of the Change

Fixes #355

Adds a paragraph discouraging the use of "EXISTS" or "NOT_EXISTS" as comparison methods to query for post meta. Please see full description as examples [in the Issue.](#355)

### Checklist:

- [x] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] All new and existing tests passed.

### Applicable Issues

#355

### DIFF
--- a/_includes/markdown/PHP.md
+++ b/_includes/markdown/PHP.md
@@ -90,9 +90,11 @@ Here are a few key points:
 
 *    [Post meta](https://wordpress.org/support/article/custom-fields/) lets us store unique information about specific posts. As such the way post meta is stored does not facilitate efficient post lookups. Generally, looking up posts by post meta should be avoided (sometimes it can't). If you have to use one, make sure that it's not the main query and that it's cached.
 
-* Always set default values for post_meta keys, especially if they are used to filter other queries. When writing queries filtering by post meta, never use compare types ```EXISTS/NOT_EXISTS```.
+* Always set default values for post_meta keys, especially if they are used to filter other queries. When writing queries filtering by post meta, never use the compare type ```NOT EXISTS```.
 
-    Queries that need to check if a given meta_key exists need to use `LEFT JOIN` and require the MySQL engine to run through all the rows in the join to make sure it catches all that meet the criteria. If all posts have a specific meta key, the engine can use a simple `INNER JOIN` and only touch the rows that already meet the criteria, saving a lot of time and processing on large tables.
+    Queries that need to check if a given meta_key does not exist need to use `LEFT JOIN`, gathering a lot more data. In practice, this type of query destroys any other performance improvements because it requires the MySQL engine to run through all the rows in the selected set to make sure it catches any rows that meet the criteria.
+    
+    If all posts have a specific meta key, or if you can test for the presence of a given key, the engine can use a simple `INNER JOIN` and only touch the rows that already meet the criteria, saving a lot of time and processing on large tables.
 
     To make sure a value was manually set on a post, use :
 
@@ -120,7 +122,7 @@ Here are a few key points:
       'meta_query' => array(
           array(
               'key' => 'meta_key',
-              'compare' => 'EXISTS'
+              'compare' => 'NOT EXISTS'
           )
       )
   ) );
--- a/_includes/markdown/PHP.md
+++ b/_includes/markdown/PHP.md
@@ -152,8 +152,6 @@ Here are a few key points:
   ?>
   ```
 
-* When multi-dimensional queries can't be avoided, Avoid using ```compare => EXISTS/NOT_EXISTS```.
-
 #### WP\_Query vs. get\_posts() vs. query\_posts()
 As outlined above, `get_posts()` and `WP_Query`, apart from some slight nuances, are quite similar. Both have the same performance cost (minus the implication of skipping filters): the query performed.
 
--- a/_includes/markdown/PHP.md
+++ b/_includes/markdown/PHP.md
@@ -88,7 +88,44 @@ Here are a few key points:
 
 * A [taxonomy](https://wordpress.org/support/article/taxonomies/) is a tool that lets us group or classify posts.
 
-    [Post meta](https://wordpress.org/support/article/custom-fields/) lets us store unique information about specific posts. As such the way post meta is stored does not facilitate efficient post lookups. Generally, looking up posts by post meta should be avoided (sometimes it can't). If you have to use one, make sure that it's not the main query and that it's cached.
+*    [Post meta](https://wordpress.org/support/article/custom-fields/) lets us store unique information about specific posts. As such the way post meta is stored does not facilitate efficient post lookups. Generally, looking up posts by post meta should be avoided (sometimes it can't). If you have to use one, make sure that it's not the main query and that it's cached.
+
+* Always set default values for post_meta keys, especially if they are used to filter other queries. When writing queries filtering by post meta, never use compare types ```EXISTS/NOT_EXISTS```.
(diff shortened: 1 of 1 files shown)


jeffpaul:
@bengreeley would appreciate review from you on this, thanks!

_includes/markdown/PHP.md
```diff
@@ -88,7 +88,46 @@ Here are a few key points:
 
 * A [taxonomy](https://wordpress.org/support/article/taxonomies/) is a tool that lets us group or classify posts.
 
-    [Post meta](https://wordpress.org/support/article/custom-fields/) lets us store unique information about specific posts. As such the way post meta is stored does not facilitate efficient post lookups. Generally, looking up posts by post meta should be avoided (sometimes it can't). If you have to use one, make sure that it's not the main query and that it's cached.
+*    [Post meta](https://wordpress.org/support/article/custom-fields/) lets us store unique information about specific posts. As such the way post meta is stored does not facilitate efficient post lookups. Generally, looking up posts by post meta should be avoided (sometimes it can't). If you have to use one, make sure that it's not the main query and that it's cached.
+
+* Always set default values for post_meta keys, especially if they are used to filter other queries. When writing queries filtering by post meta, never use the compare type ```NOT EXISTS```.
```
bengreeley, line 93: @moraleida Could you put an example of when we'd set default values for post_meta keys? Is that when using update_post_meta or add_post_meta? I feel like we need a bit more information on why default values (instead of empty strings, I assume) are a best practice.

bengreeley: I'd recommend we lead with 'Never use compare type NOT EXISTS' as the leading part of this (or on its own line) since it tees up the majority of the rest of the section

_includes/markdown/PHP.md
```diff
@@ -88,7 +88,46 @@ Here are a few key points:
...
+
+    Queries that need to check if a given meta_key does not exist need to use `LEFT JOIN`, gathering a lot more data. In practice, this type of query destroys any other performance improvements because it requires the MySQL engine to run through all the rows in the selected set to make sure it catches any rows that meet the criteria.
+    
+    If all posts have a specific meta key, or if you can test for the presence of a given key, the engine can use a simple `INNER JOIN` and only touch the rows that already meet the criteria, saving a lot of time and processing on large tables.
```
bengreeley, line 97: I'd recommend rewording `If all posts have a specific meta key, or if you can test for the presence of a given key` to `To search for the presence of a given key`. The 'if all posts have a specific meta key' doesn't make a ton of sense to me, but I could be missing something.

bengreeley (changes requested):
@moraleida Thank you for putting this together! I made a few suggestions in wording that will help clarify the intent and really focus on the performance recommendations. Let me know if you have any questions.

PR #378, opened 2021-11-05
Updated ENV_DEVELOPMENT to WP_ENVIRONMENT_TYPE.

jamesmorrison:
### Description of the Change

This PR updates the `ENV_DEVELOPMENT` constant to `WP_ENVIRONMENT_TYPE`. `WP_ENVIRONMENT_TYPE` was introduced in WordPress 5.5 and standardises both the constant and expected values [`production` / `staging` / `development` / `local`] to determine the current environment.

### Alternate Designs

This is a WordPress core feature, no alternative designs were considered.

### Benefits

The key benefit is standardisation of the constant name and expected values.

### Possible Drawbacks

Any projects already reliant on `ENV_DEVELOPMENT` (and possibly corresponding `ENV_STAGING` / `ENV_PRODUCTION` should continue to do so, or update the constant across all environments.

### Verification Process

This change in the documentation was tested locally with `bundle exec jekyll serve`.

### Checklist:

- [ ] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] All new and existing tests passed.

### Applicable Issues

None - but I can create one if required.

### Changelog Entry

CHANGED: Updated `ENV_DEVELOPMENT` to `WP_ENVIRONMENT_TYPE`

### DIFF
--- a/_includes/markdown/Design.md
+++ b/_includes/markdown/Design.md
@@ -172,13 +172,13 @@ If a management UI is impossible due to the nature of the project, credentials s
 // Production API keys should ideally be defined in wp-config.php
 // This section should default to a development or noop key instead.
 if ( ! defined( 'CLIENT_MANDRILL_API_KEY' ) {
-    if ( defined( 'WP_ENVIRONMENT_TYPE' ) && 'production' === wp_get_environment_type() ) {
+    if ( 'production' === wp_get_environment_type() ) {
         define( 'CLIENT_MANDRILL_API_KEY', '1234567890' );
     }
 }
 ```
 
-The `WP_ENVIRONMENT_TYPE` constant should always be set to `local` or `development` for local development and should be used whenever and wherever possible to prevent production-only functionality from triggering in a local environment.
+The `WP_ENVIRONMENT_TYPE` constant should always be set to `local` or `development` for local development and should be used whenever and wherever possible to prevent production-only functionality from triggering in a non-production environment.
 
 For futher reading on WordPress environment types, see the function documentation [wp_get_environment_type()](https://developer.wordpress.org/reference/functions/wp_get_environment_type/) and the introduction post [New wp_get_environment_type() function in WordPress 5.5](https://make.wordpress.org/core/2020/07/24/new-wp_get_environment_type-function-in-wordpress-5-5/).
 
--- a/_includes/markdown/Design.md
+++ b/_includes/markdown/Design.md
@@ -171,12 +171,16 @@ If a management UI is impossible due to the nature of the project, credentials s
 <?php
 // Production API keys should ideally be defined in wp-config.php
 // This section should default to a development or noop key instead.
-if ( ! defined( 'CLIENT_MANDRILL_API_KEY' ) && ! ENV_DEVELOPMENT ) {
+if ( ! defined( 'CLIENT_MANDRILL_API_KEY' ) && defined( 'WP_ENVIRONMENT_TYPE' ) && ! in_array( wp_get_environment_type(), array( 'local', 'development' ) ) ) {
 	define( 'CLIENT_MANDRILL_API_KEY', '1234567890' );
 }
 ```
 
-The `ENV_DEVELOPMENT` constant should always be set to `true` for local development and should be used whenever and wherever possible to prevent production-only functionality from triggering in a local environment.
+The `WP_ENVIRONMENT_TYPE` constant should always be set to `local` or `development` for local development and should be used whenever and wherever possible to prevent production-only functionality from triggering in a local environment.
+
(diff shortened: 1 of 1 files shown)


_includes/markdown/Design.md
```diff
@@ -171,12 +171,16 @@ If a management UI is impossible due to the nature of the project, credentials s
 <?php
 // Production API keys should ideally be defined in wp-config.php
 // This section should default to a development or noop key instead.
-if ( ! defined( 'CLIENT_MANDRILL_API_KEY' ) && ! ENV_DEVELOPMENT ) {
+if ( ! defined( 'CLIENT_MANDRILL_API_KEY' ) && defined( 'WP_ENVIRONMENT_TYPE' ) && ! in_array( wp_get_environment_type(), array( 'local', 'development' ) ) ) {
```
gthayer: @jamesmorrison - I think we should be more restrictive and only include the keys when it's production, instead of avoiding it on locals.

```
if ( ! defined( 'CLIENT_MANDRILL_API_KEY' ) {
    if ( defined( 'WP_ENVIRONMENT_TYPE' ) && 'production' === wp_get_environment_type() ) {
        define( 'CLIENT_MANDRILL_API_KEY', '1234567890' );
    }
}
```

My reasoning for this is that staging/preprod environments would include production creds otherwise.

gthayer (changes requested):
@jamesmorrison - Thanks for putting this together. I have a slight recommendation around the code example, but the write up is perfect.

_includes/markdown/Design.md
gthayer: Also, you don't need to check for `defined( 'WP_ENVIRONMENT_TYPE' )`.

`wp_get_environment_type()` does this for you, and returns `''` if it's not set.

jamesmorrison: I've updated the example per your suggestion, in https://github.com/10up/Engineering-Best-Practices/pull/378/commits/7f88bcb24f3e1bbd8a4e7875a6e60eaae6bdd6f3

Let me know if you think this should be expanded on - for example to include examples for other environments, or via a `switch` statement (based on the value returned from `wp_get_environment_type`).

_includes/markdown/Design.md
```diff
@@ -171,12 +171,18 @@ If a management UI is impossible due to the nature of the project, credentials s
 <?php
 // Production API keys should ideally be defined in wp-config.php
 // This section should default to a development or noop key instead.
-if ( ! defined( 'CLIENT_MANDRILL_API_KEY' ) && ! ENV_DEVELOPMENT ) {
-	define( 'CLIENT_MANDRILL_API_KEY', '1234567890' );
+if ( ! defined( 'CLIENT_MANDRILL_API_KEY' ) {
+    if ( defined( 'WP_ENVIRONMENT_TYPE' ) && 'production' === wp_get_environment_type() ) {
```
gthayer: @jamesmorrison - This is picky, but you don't need `defined( 'WP_ENVIRONMENT_TYPE'  ) && `.

`wp_get_environment_type` already includes this check and if the value is not set it returns `''`

_includes/markdown/Design.md
```diff
@@ -171,12 +171,18 @@ If a management UI is impossible due to the nature of the project, credentials s
...
+        define( 'CLIENT_MANDRILL_API_KEY', '1234567890' );
+    }
 }
 ```
 
-The `ENV_DEVELOPMENT` constant should always be set to `true` for local development and should be used whenever and wherever possible to prevent production-only functionality from triggering in a local environment.
+The `WP_ENVIRONMENT_TYPE` constant should always be set to `local` or `development` for local development and should be used whenever and wherever possible to prevent production-only functionality from triggering in a local environment.
```
gthayer: Recommend changing `local environment` to `non-production`.

gthayer (changes requested):
@jamesmorrison - A couple a picky minor notes for your review. Thanks for keeping this moving!

_includes/markdown/Design.md
jamesmorrison: @gthayer This has been updated from `local` => `non-production`

_includes/markdown/Design.md
jamesmorrison: @gthayer I removed the `defined( 'WP_ENVIRONMENT_TYPE' )` conditional.

_includes/markdown/Design.md
jamesmorrison: @gthayer This has been updated.

gthayer:
LGTM!

PR #403, opened 2024-05-22
Introducing TypeScript best practices

nicholasio:
This PR introduces a new TypeScript section and its best practices.

### DIFF
--- a/_includes/markdown/Typescript.md
+++ b/_includes/markdown/Typescript.md
@@ -1,0 +1,73 @@
+The purpose of this document is to help guide Engineers at the usage of TypeScript at 10up projects. The goal here is not to teach TypeScript but rather explain 10up recommendation on when to use TypeScript alongside general recommendations.
+
+## Why TypeScript
+
+We believe that:
+1. TypeScript can make engineers more confident
+2. TypeScript makes refactoring code easier
+3. TypeScript helps engineers unfamiliar with the codebase
+
+Therefore we are big fans of TypeScript at 10up and we strive to use whenever it makes sense.
+
+## When to use TypeScript
+
+At 10up we require that all projects considered "JavaScript-First" to make use of TypeScript. Generally speaking Standard WordPress projects (including block development) would not fall under "JavaScript-First" projects. 
+
+If you are building a project using Next.js and/or 10up's [HeadstartWP framework](https://headstartwp.10up.com/), Sanity and any other headless CMS'es powered by a React front-end then we require TypeScript unless there are other cicunstances that makes using just JavaScript a better choice.
+
+The main reason we don't require TypeScript for standard WordPress projects at the moment, is because TS support for block development is still very limited, largely driven by community types packages that don't always match the actual types for the `@wordpress` packages.
+
+## How to use TypeScript
+
+In an ideal world every engineer would be fully capable in writing all sorts of complex TypeScript types, however we understand that it takes time to master a language like TypeScript, specially in a few scenarios that requires writing more complex types. Therefore, while we use a reasonable strict `tsconfig.json` we are flexible in practice. 
+
+We want the TypeScript to always be valid and issue no errors at compilation, but we allow a few escape hatches should engineers get stuck with a particular typing problem. We follow the principles below when writing TypeScript:
+
+### 1. Never ship code that has compilation issues
+
+Depending on the tooling you're using (e.g Next.js) it is possible to disable type checking and still compile invalid TS code as long as the underlying JS code is valid. While the runtime behavior might still be working as expected this not a recommended practice and generally would just make engineers to not have trust in the type system. If you are facing an challenge typing something it is better to use some of the escape hatches documented below instead of just ignoring TS errors.
+
+### 2. Favor efficiency instead of strict typing
+
+In ideal world all engineers would be capable of typing everything properly from the get go. However we know that is not the case and engineers needs time and practice to get there. For that reason we favor efficiency over strict typing. 
+
+Our philosophy is that over time engineers will organically get more familiar with TypeScript and stricness will come as a result. Code reviews are also a place where engineers might get quick feedback that would allow them to more efficiently adress typing issues instead of sinking a lot of time during initial development trying to figure it out on their own.
+
+When facing such typing challenges we recommend the following strategies as a escape hatch:
+#### 1. The `as unkown as ExpectedType` type cast
+
+When truggling with a type that is not matching the expected type, one of the escape hatches we recommend is to forcebly type cast the type to the expected type:
+
+```ts
+// force the variable to have the type `ExpectedType`
+const withExpectedType = myValueThatWithWrongType as unkown as ExpectedType;
+```
+
+While this usually indicates that there's something wrong with types or even the code itself, this would at least provide type checking for the subsequent lines of code. This strategy for instance, is much better than just typing as any.
+
+#### 2. The `any` type.
+
+Another strategy is to just type your variable with `any` which essentially disables type checking for that variable and while it is a more drastic measure it should be an easy way to unblock work until engineers have more time to come back and try to get the types right.
+
+Note: we recommend setting the `noImplicitAny` setting to `true` in `tsconfig.json` which forces explicitly setting a type as any. This makes it easier to identify code that is relying on any and would also make catching the escape hatch in code reviews which is a great opportunity to get feedback and potentially get the types right.
+
+#### 3. The `// @ts-expect-error` comment
+
+Latsly, if nothing else worked to unblock the code, TypeScript allows disregarding errors on a given line by just telling the compiler that an error is expected on that line. This is usually the least effective option as it would only solve the problem for a single line. However, if nothing else works, this is the last and most drastic escape hatch.
+
+While we'd like that none of these escape hatches to be used, it would be unrealistic to expect all engineers to have types 100% correct on every project all the time. We believe that these recomendations provides a path to address them organically during the life-cycle of the project.
+
+## TypeScript and CI
+
+We recommend that all projects run the typescript compiler to check type safety at the CI level. Every Pull Request/Merge Request should run `tsc --noEmit` on CI to ensure there are no compilation issues before merging the branch.
+
+This is a great strategy to avoid builds failing due to type errors. With Next.js for instance, you might not realise you have a TS compilation issue until you actually run a build, therefore to avoid merging code that produces failing builds we recommend always ensuring that the code has no TypeScript compilation issues before merging.
+
+## Recommended `tsconfig.json` settings
+
+This is not a full list but we highlight some recommended settings here.
+
+1. `noFallthroughCasesInSwitch: true`- [Definition](https://www.typescriptlang.org/tsconfig/#noFallthroughCasesInSwitch) - ensures you are not skipping anything in a switch statement, and always at least provide a default option.
+2. `noImplicitAny: true` - [Definition](https://www.typescriptlang.org/tsconfig/#noImplicitAny) - ensures you define `any` explicitly, rather than implicitly assume it. This makes it a lot easier to find all code relying on any for future refactoring.
+3. `strictNullChecks: true` - [Definition](https://www.typescriptlang.org/tsconfig/#strictNullChecks) - forces handling both `undefined` and `null`
+4. `noUnusedParameters: true` - [Definition](https://www.typescriptlang.org/tsconfig/#noUnusedParameters) - does not allow unused parameters to remain in code
--- a/_includes/markdown/Typescript.md
+++ b/_includes/markdown/Typescript.md
@@ -138,7 +138,7 @@ const PostProps = {
 const Post = (props: PostProps) => {
     return (
         <article>
-            <PostContent {...props} >
+            <PostContent {...props}>
         </article>
     );
 }
--- a/_includes/markdown/Typescript.md
+++ b/_includes/markdown/Typescript.md
@@ -6,33 +6,41 @@ We believe that:
 1. TypeScript can make engineers more confident
 2. TypeScript makes refactoring code easier
 3. TypeScript helps engineers unfamiliar with the codebase
+4. TypeScript helps catch certain issues earlier
 
 Therefore we are big fans of TypeScript at 10up and we strive to use it whenever it makes sense.
+
+We also recgonize that the industry has largely adopted TypeScript as the defacto standard for writing JavaScript apps.
 
 ## When to use TypeScript
 
 At 10up we require that all projects considered "JavaScript-First" to make use of TypeScript. Generally speaking Standard WordPress projects (including block development) would not fall under "JavaScript-First" projects. 
 
-If you are building a project using Next.js and/or 10up's [HeadstartWP framework](https://headstartwp.10up.com/), Sanity and any other headless CMS'es powered by a React front-end then we require TypeScript unless other circumstances make using just JavaScript a better choice.
+If you are building a project using Next.js and/or 10up's [HeadstartWP framework](https://headstartwp.10up.com/), Sanity and any other headless CMS'es powered by a React front-end then we require TypeScript unless other circumstances make using JavaScript a better choice.
 
-The main reason we don't require TypeScript for standard WordPress projects at the moment is that TypeScript support for block development is still very limited, largely driven by community types packages that don't always match the actual types for the `@wordpress` packages.
+The main reason we don't require TypeScript for standard WordPress projects is that TypeScript support for block development is still very limited, largely driven by community types packages that don't always match the actual types for the `@wordpress` packages.
 
 ## How to use TypeScript
 
-In an ideal world, every engineer would be fully capable of writing all sorts of complex TypeScript types, however, we understand that it takes time to master a language like TypeScript, especially in a few scenarios that require writing more complex types. Therefore, while we use a reasonably strict `tsconfig.json` we are flexible in practice.
+In an ideal world, no one would struggle with TypeScript, however, we understand that it takes time to master a language like TypeScript, especially in a few scenarios that require writing more complex types. Therefore, while we use a reasonably strict `tsconfig.json` we are flexible in practice.
 
-We want the TypeScript to always be valid and issue no errors at compilation, but we allow a few escape hatches should engineers get stuck with a particular typing problem. We follow the principles below when writing TypeScript:
+We also don't want engineers looking for "perfection" when it comes to writing complex types. TypeScript has a very powerfull type system and it allows for amazing things to be done statically at compilation level, however, we should recognize that most projects will suffice with simpler, easy-to-use types. Unless there's a good reason (e.g building a reusable package or library), engineers should strive to not write overly complex types.
+
+Lastly, we want TypeScript to always be valid and issue no errors at compilation, but we allow a few escape hatches should engineers get stuck with a particular typing problem. We follow the principles below when writing TypeScript:
 
 ### 1. Never ship code that has compilation issues
 
 Depending on the tooling you're using (e.g. Next.js) it is possible to disable type checking and still compile invalid TS code as long as the underlying JS code is valid. While the runtime behavior might still be working as expected this is not a recommended practice and generally would just make engineers not have trust in the type system. If you are facing a challenge typing something it is better to use some of the escape hatches documented below instead of just ignoring TS errors.
 
 ### 2. Favor efficiency instead of strict typing
+
 In an ideal world, all engineers would be capable of typing everything properly from the get-go. However, we know that is not the case and engineers need time and practice to get there. For that reason, we favor efficiency over strict typing. 
+
+By "efficiency" we mean engineers recognizing and time-boxing the time spent trying to type something properly and not letting that block the development of the task at hand. We prefer to ship non-strict typed code instead of sinking a lot of time on it.
 
 Our philosophy is that over time engineers will organically get more familiar with TypeScript and strictness will come as a result. Code reviews are also a place where engineers might get quick feedback that would allow them to more efficiently address typing issues instead of wasting a lot of time during initial development trying to figure it out on their own.
 
-When facing such typing challenges we recommend the following strategies as an escape hatch:
+To favor efficiency, we recommend the following "escape hatches":
 
 #### 1. The `as unknown as ExpectedType` type cast
 
@@ -55,12 +63,18 @@ Note: we recommend setting the `noImplicitAny` setting to `true` in `tsconfig.js
 
 Lastly, if nothing else worked to unblock the code, TypeScript allows disregarding errors on a given line by just telling the compiler that an error is expected on that line. This is usually the least effective option as it would only solve the problem for a single line. However, if nothing else works, this is the last and most drastic escape hatch.
 
+When using any of these escape hatches, please leave a `TODO` comment with a short description of the issue:
+
+```js
+// TODO(nicholaiso): the types coming from the package do not seem correct
+```
+
 While we'd like none of these escape hatches to be used, it would be unrealistic to expect all engineers to have types 100% correct on every project all the time. We believe that these recommendations provide a path to address them organically during the life-cycle of the project.
 
 ## Tools
 
 - [TypeScript Error Translator](https://ts-error-translator.vercel.app/): Lets you drop any TypeScript error message and gives you a human-readable explanation in plain English.
-- [pretty-ts-errors] (https://marketplace.visualstudio.com/items?itemName=yoavbls.pretty-ts-errors): A VSCode extension that translaters TypeScript errors in a human-readable way right into VSCode.
+- [pretty-ts-errors](https://marketplace.visualstudio.com/items?itemName=yoavbls.pretty-ts-errors): A VSCode extension that translaters TypeScript errors in a human-readable way right into VSCode.
 
 ## TypeScript and CI
 
@@ -85,21 +99,40 @@ We also have a recommended [tsconfig.json](https://github.com/10up/headstartwp/b
 
 While `interface` and `type` are mostly interchangeable, prefer `type` over `interfaces` unless you expect the props to be the base for other components's props.
 
-```ts
+```tsx
 type BlocksProps = {
 	html: string;
 };
 
-export const Blocks = ({ html }: BlocksProps) => {
+export const Blocks: React.FC<BlockProps> = ({ html }) => {
     // render blocks
 }
 ```
+
+### Use `React.FC` to type React Components
+
+When typing React components prefer using `React.FC`. While in the past it was problematic because of implicitly accepting `children` as an optional prop, since React 18 it [no longer does that](https://www.totaltypescript.com/you-can-stop-hating-react-fc). Using React.FC ensures your component is correctly typed, including its return type.
+
+```tsx
+import * as React from 'react';
+
+type MyComponentProps = {
+	/* */
+};
+
+export const MyComponent: React.FC<MyComponentProps> = (props) => {
+	// component logic
+}
+```
+**Note**: If using React < 17, be aware that `children` will be explicitly set for all components using `React.FC`. This can cause some TS errors when upgrading React to 18+, if a component is passed `children` but does not explicitly declare it as a prop.
+
+The [React TypeScript Cheatsheet](https://react-typescript-cheatsheet.netlify.app/) is a good resource to check how to type specific things in React (e.g Forms, Events etc).
 
 ### Use function default arguments instead of `defaultProps`
 
 `defaultProps` has been deprecated in React 19, therefore prefer using function default arguments.
 
-```ts
+```tsx
 type MyComponentProps = {
     title?: string;
 };
@@ -115,12 +148,12 @@ Make sure the eslint rule `react/require-default-props` is set up to look for `d
 
 `PropsWithChildren` will make `children` optional, but sometimes you do want to be very explicit and make it required. Therefore we generally recommend declaring `children` explicitly.
 
-```ts
+```tsx
 type LayoutProps = {
-    children: React.Node
+    children: React.Node;
 };
 
-// Layout has passed a children prop
+// If Layout is not passed a children, TS will catch the error
 const Layout = (props: LayoutProps) {
     return (
         <main>
@@ -134,7 +167,7 @@ const Layout = (props: LayoutProps) {
 
 Prop spreading is when you simply forward all props to another component e.g:
 
-```ts
+```tsx
 // AVOID DOING THIS
 const PostProps = {
     title: string,
--- a/_includes/markdown/Typescript.md
+++ b/_includes/markdown/Typescript.md
@@ -24,7 +24,7 @@ The main reason we don't require TypeScript for standard WordPress projects is t
 
 In an ideal world, no one would struggle with TypeScript, however, we understand that it takes time to master a language like TypeScript, especially in a few scenarios that require writing more complex types. Therefore, while we use a reasonably strict `tsconfig.json` we are flexible in practice.
 
-We also don't want engineers looking for "perfection" when it comes to writing complex types. TypeScript has a very powerfull type system and it allows for amazing things to be done statically at compilation level, however, we should recognize that most projects will suffice with simpler, easy-to-use types. Unless there's a good reason (e.g building a reusable package or library), engineers should strive to not write overly complex types.
+We also don't want engineers looking for "perfection" when it comes to writing complex types. TypeScript has a very powerful type system and it allows for amazing things to be done statically at the compilation level, however, we should recognize that most projects will suffice with simpler, easy-to-use types. Unless there's a good reason (e.g. building a reusable package or library), engineers should strive to not write overly complex types.
 
 Lastly, we want TypeScript to always be valid and issue no errors at compilation, but we allow a few escape hatches should engineers get stuck with a particular typing problem. We follow the principles below when writing TypeScript:
 
--- a/_includes/markdown/Typescript.md
+++ b/_includes/markdown/Typescript.md
@@ -142,7 +142,7 @@ export const MyComponent = ({ title = 'Default Title' }: MyComponentsProps) => {
 }
 ```
 
-Make sure the eslint rule `react/require-default-props` is set up to look for `defaultArguments` e.g `'react/require-default-props': ['error', { functions: 'defaultArguments' }]`;
+Make sure the ESLint rule `react/require-default-props` is set up to look for `defaultArguments` e.g `'react/require-default-props': ['error', { functions: 'defaultArguments' }]`;
 
 ### Prefer explicitly declaring `children` instead of using `PropsWithChildren`
 
--- a/_includes/markdown/Typescript.md
+++ b/_includes/markdown/Typescript.md
@@ -126,7 +126,7 @@ export const MyComponent: React.FC<MyComponentProps> = (props) => {
 ```
 **Note**: If using React < 17, be aware that `children` will be explicitly set for all components using `React.FC`. This can cause some TS errors when upgrading React to 18+, if a component is passed `children` but does not explicitly declare it as a prop.
 
-The [React TypeScript Cheatsheet](https://react-typescript-cheatsheet.netlify.app/) is a good resource to check how to type specific things in React (e.g Forms, Events etc).
+The [React TypeScript Cheatsheet](https://react-typescript-cheatsheet.netlify.app/) is a good resource to check how to type specific things in React (e.g. Forms, Events etc).
 
 ### Use function default arguments instead of `defaultProps`
 
--- a/_includes/markdown/Typescript.md
+++ b/_includes/markdown/Typescript.md
@@ -16,7 +16,7 @@ We also recognize that the industry has largely adopted TypeScript as the de fac
 
 At 10up we require that all projects considered "JavaScript-First" to make use of TypeScript. Generally speaking Standard WordPress projects (including block development) would not fall under "JavaScript-First" projects. 
 
-If you are building a project using Next.js and/or 10up's [HeadstartWP framework](https://headstartwp.10up.com/), Sanity and any other headless CMS'es powered by a React front-end then we require TypeScript unless other circumstances make using JavaScript a better choice.
+If you are building a project using Next.js and/or 10up's [HeadstartWP framework](https://headstartwp.10up.com/), Sanity and any other headless CMSs' powered by a React front-end then we require TypeScript unless other circumstances make using JavaScript a better choice.
 
 The main reason we don't require TypeScript for standard WordPress projects is that TypeScript support for block development is still very limited, largely driven by community types packages that don't always match the actual types for the `@wordpress` packages.
 
--- a/_includes/markdown/Typescript.md
+++ b/_includes/markdown/Typescript.md
@@ -70,3 +70,111 @@ This is not a full list but we highlight some recommended settings here.
 2. `noImplicitAny: true` - [Definition](https://www.typescriptlang.org/tsconfig/#noImplicitAny) - ensures you define `any` explicitly, rather than implicitly assume it. This makes it a lot easier to find all code relying on any for future refactoring.
 3. `strictNullChecks: true` - [Definition](https://www.typescriptlang.org/tsconfig/#strictNullChecks) - forces handling both `undefined` and `null`
 4. `noUnusedParameters: true` - [Definition](https://www.typescriptlang.org/tsconfig/#noUnusedParameters) - does not allow unused parameters to remain in code
+
+## TypeScript and React
+
+### Use `type` for React Props
+
+While `interface` and `type` are mostly interchangeable, prefer `type` over `interfaces` unless you expect the props to be the base for other components's props.
+
+```ts
+type BlocksProps = {
+	html: string;
+};
+
+export const Blocks = ({ html }: BlocksProps) => {
+    // render blocks
+}
+```
+
+### Use function default arguments instead of `defaultProps`
+
+`defaultProps` has been deprecated in React 19, therefore prefer using function default arguments.
+
+```ts
+type MyComponentProps = {
+    title?: strinmg;
+};
+
+export const MyComponent = ({ title = 'Default Title' }: MyComponentsProps) => {
+    // ...
+}
+```
+
+Make sure the eslint rule `react/require-default-props` is set up to look for `defaultArguments` e.g `'react/require-default-props': ['error', { functions: 'defaultArguments' }]`;
+
+### Prefer explicitly declaring `children` instead of using `PropsWithChildren`
+
+`PropsWithChildren` will make `children` optional, but sometimes you do want to be very explicit and make it required. Therefore we generally recommend declaring `children` explicitly.
+
+```ts
+type LayoutProps = {
+    children: React.Node
+};
+
+// Layout has passed a children prop
+const Layout = (props: LayoutProps) {
+    return (
+        <main>
+            {children}
+        </main>
+    );
+}
+```
+
+### Avoid prop spreading
+
+Prop spreading is when you simply forward all props to another component e.g:
+
+```ts
+// AVOID DOING THIS
+const PostProps = {
+    title: string,
+    category: string,
+    link: string
+    // other props
+};
+
+const Post = (props: PostProps) => {
+    return (
+        <article>
+            <PostContent {...props} >
+        </article>
+    );
+}
+```
+
+While there might be use cases where you need to use them, it's generally better to avoid them and pass props explicitly. In the example above it would be better to just pass a `post` object as a single prop.
+
+Prop spreading is especially dangerous when prop spreading into native DOM elements. So make sure to not ignore TypeScript errors and make sure your types are set up correctly so that TS can catch forwarded props incompatible with the target component/element.
+
+### Colocate types with React components
+
+For the most part, the React Component types should be colocated with the React component, meaning that it should be in the same file that the React component is written. Only hoist types to avoid circular dependencies and when you expect them to be reused/shared across many components/files.
+
+### Use `@types` packages if necessary
+
+If the library you're using does not ship types, check if there is type information available for that package in the [DefinatelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) repo.
+
+### Global types definitions
+
+**Do not** put the application types in the global scope just to avoid importing them. You may only create global type definitions when needing to extend global types (e.g `window` object).
+
+To extend global types create a `global.d.ts` file and include them in your `tsconfig.json` in the `include` config option. Here's an example for a Next.js project:
+
+```json
+ "include": [
+    "next-env.d.ts",
+    "src/global.d.ts",
+    "**/*.ts",
+    "**/*.tsx"
+  ],
+```
+
+```ts
+// globals.d.ts
+interface Window {
+    // add any methods or props added by third party scripts/libraries
+    custom_prop: number;
+}
+```
--- a/_includes/markdown/Typescript.md
+++ b/_includes/markdown/Typescript.md
@@ -33,6 +33,7 @@ In an ideal world, all engineers would be capable of typing everything properly 
 Our philosophy is that over time engineers will organically get more familiar with TypeScript and strictness will come as a result. Code reviews are also a place where engineers might get quick feedback that would allow them to more efficiently address typing issues instead of wasting a lot of time during initial development trying to figure it out on their own.
 
 When facing such typing challenges we recommend the following strategies as an escape hatch:
+
 #### 1. The `as unknown as ExpectedType` type cast
 
 When struggling with a type that does not match the expected type, one of the escape hatches we recommend is to forcibly typecast the type to the expected type:
@@ -70,6 +71,8 @@ This is not a full list but we highlight some recommended settings here.
 2. `noImplicitAny: true` - [Definition](https://www.typescriptlang.org/tsconfig/#noImplicitAny) - ensures you define `any` explicitly, rather than implicitly assume it. This makes it a lot easier to find all code relying on any for future refactoring.
 3. `strictNullChecks: true` - [Definition](https://www.typescriptlang.org/tsconfig/#strictNullChecks) - forces handling both `undefined` and `null`
 4. `noUnusedParameters: true` - [Definition](https://www.typescriptlang.org/tsconfig/#noUnusedParameters) - does not allow unused parameters to remain in code
+
+We also have a recommended [tsconfig.json](https://github.com/10up/headstartwp/blob/develop/projects/wp-nextjs/tsconfig.json) for Next.js projects.
 
 ## TypeScript and React
 
--- a/_includes/markdown/Typescript.md
+++ b/_includes/markdown/Typescript.md
@@ -199,7 +199,7 @@ If the library you're using does not ship types, check if there is type informat
 
 ### Global types definitions
 
-**Do not** put the application types in the global scope just to avoid importing them. You may only create global type definitions when needing to extend global types (e.g `window` object).
+**Do not** put the application types in the global scope just to avoid importing them. You may only create global type definitions when needing to extend global types (e.g. `window` object).
 
 To extend global types create a `global.d.ts` file and include them in your `tsconfig.json` in the `include` config option. Here's an example for a Next.js project:
 
--- a/_includes/markdown/Typescript.md
+++ b/_includes/markdown/Typescript.md
@@ -10,7 +10,7 @@ We believe that:
 
 Therefore we are big fans of TypeScript at 10up and we strive to use it whenever it makes sense.
 
-We also recgonize that the industry has largely adopted TypeScript as the defacto standard for writing JavaScript apps.
+We also recognize that the industry has largely adopted TypeScript as the de facto standard for writing JavaScript apps.
 
 ## When to use TypeScript
 
--- a/_includes/markdown/Typescript.md
+++ b/_includes/markdown/Typescript.md
@@ -101,7 +101,7 @@ export const Blocks = ({ html }: BlocksProps) => {
 
 ```ts
 type MyComponentProps = {
-    title?: strinmg;
+    title?: string;
 };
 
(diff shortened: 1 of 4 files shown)


_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,73 @@
...
+
+Depending on the tooling you're using (e.g Next.js) it is possible to disable type checking and still compile invalid TS code as long as the underlying JS code is valid. While the runtime behavior might still be working as expected this not a recommended practice and generally would just make engineers to not have trust in the type system. If you are facing an challenge typing something it is better to use some of the escape hatches documented below instead of just ignoring TS errors.
+
+### 2. Favor efficiency instead of strict typing
+
+In ideal world all engineers would be capable of typing everything properly from the get go. However we know that is not the case and engineers needs time and practice to get there. For that reason we favor efficiency over strict typing. 
+
+Our philosophy is that over time engineers will organically get more familiar with TypeScript and stricness will come as a result. Code reviews are also a place where engineers might get quick feedback that would allow them to more efficiently adress typing issues instead of sinking a lot of time during initial development trying to figure it out on their own.
```
pattonwebz:
```suggestion
Our philosophy is that over time engineers will organically get more familiar with TypeScript and strictness will come as a result. Code reviews are also a place where engineers might get quick feedback that would allow them to more efficiently address typing issues instead of sinking a lot of time during initial development trying to figure it out on their own.
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,73 @@
...
+
+When facing such typing challenges we recommend the following strategies as a escape hatch:
+#### 1. The `as unkown as ExpectedType` type cast
+
+When truggling with a type that is not matching the expected type, one of the escape hatches we recommend is to forcebly type cast the type to the expected type:
+
+```ts
+// force the variable to have the type `ExpectedType`
+const withExpectedType = myValueThatWithWrongType as unkown as ExpectedType;
```
pattonwebz:
```suggestion
const withExpectedType = myValueThatWithWrongType as unknown as ExpectedType;
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,73 @@
...
+2. TypeScript makes refactoring code easier
+3. TypeScript helps engineers unfamiliar with the codebase
+
+Therefore we are big fans of TypeScript at 10up and we strive to use whenever it makes sense.
+
+## When to use TypeScript
+
+At 10up we require that all projects considered "JavaScript-First" to make use of TypeScript. Generally speaking Standard WordPress projects (including block development) would not fall under "JavaScript-First" projects. 
+
+If you are building a project using Next.js and/or 10up's [HeadstartWP framework](https://headstartwp.10up.com/), Sanity and any other headless CMS'es powered by a React front-end then we require TypeScript unless there are other cicunstances that makes using just JavaScript a better choice.
```
pattonwebz:
```suggestion
If you are building a project using Next.js and/or 10up's [HeadstartWP framework](https://headstartwp.10up.com/), Sanity and any other headless CMS'es powered by a React front-end then we require TypeScript unless there are other circumstances that makes using just JavaScript a better choice.
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,73 @@
...
+### 2. Favor efficiency instead of strict typing
+
+In ideal world all engineers would be capable of typing everything properly from the get go. However we know that is not the case and engineers needs time and practice to get there. For that reason we favor efficiency over strict typing. 
+
+Our philosophy is that over time engineers will organically get more familiar with TypeScript and stricness will come as a result. Code reviews are also a place where engineers might get quick feedback that would allow them to more efficiently adress typing issues instead of sinking a lot of time during initial development trying to figure it out on their own.
+
+When facing such typing challenges we recommend the following strategies as a escape hatch:
+#### 1. The `as unkown as ExpectedType` type cast
+
+When truggling with a type that is not matching the expected type, one of the escape hatches we recommend is to forcebly type cast the type to the expected type:
```
pattonwebz:
```suggestion
When struggling with a type that is not matching the expected type, one of the escape hatches we recommend is to forcibly type cast the type to the expected type:
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,73 @@
...
+
+#### 2. The `any` type.
+
+Another strategy is to just type your variable with `any` which essentially disables type checking for that variable and while it is a more drastic measure it should be an easy way to unblock work until engineers have more time to come back and try to get the types right.
+
+Note: we recommend setting the `noImplicitAny` setting to `true` in `tsconfig.json` which forces explicitly setting a type as any. This makes it easier to identify code that is relying on any and would also make catching the escape hatch in code reviews which is a great opportunity to get feedback and potentially get the types right.
+
+#### 3. The `// @ts-expect-error` comment
+
+Latsly, if nothing else worked to unblock the code, TypeScript allows disregarding errors on a given line by just telling the compiler that an error is expected on that line. This is usually the least effective option as it would only solve the problem for a single line. However, if nothing else works, this is the last and most drastic escape hatch.
```
pattonwebz:
```suggestion
Lastly, if nothing else worked to unblock the code, TypeScript allows disregarding errors on a given line by just telling the compiler that an error is expected on that line. This is usually the least effective option as it would only solve the problem for a single line. However, if nothing else works, this is the last and most drastic escape hatch.
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,73 @@
...
+
+While we'd like that none of these escape hatches to be used, it would be unrealistic to expect all engineers to have types 100% correct on every project all the time. We believe that these recomendations provides a path to address them organically during the life-cycle of the project.
```
pattonwebz:
```suggestion
While we'd like that none of these escape hatches to be used, it would be unrealistic to expect all engineers to have types 100% correct on every project all the time. We believe that these recommendations provides a path to address them organically during the life-cycle of the project.
```

pattonwebz (commented):
I spotted some typos in the new docs that I created suggestions for.

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,73 @@
...
+
+## How to use TypeScript
+
+In an ideal world every engineer would be fully capable in writing all sorts of complex TypeScript types, however we understand that it takes time to master a language like TypeScript, specially in a few scenarios that requires writing more complex types. Therefore, while we use a reasonable strict `tsconfig.json` we are flexible in practice. 
+
+We want the TypeScript to always be valid and issue no errors at compilation, but we allow a few escape hatches should engineers get stuck with a particular typing problem. We follow the principles below when writing TypeScript:
+
+### 1. Never ship code that has compilation issues
+
+Depending on the tooling you're using (e.g Next.js) it is possible to disable type checking and still compile invalid TS code as long as the underlying JS code is valid. While the runtime behavior might still be working as expected this not a recommended practice and generally would just make engineers to not have trust in the type system. If you are facing an challenge typing something it is better to use some of the escape hatches documented below instead of just ignoring TS errors.
```
pattonwebz:
```suggestion
Depending on the tooling you're using (e.g. Next.js) it is possible to disable type checking and still compile invalid TS code as long as the underlying JS code is valid. While the runtime behavior might still be working as expected this not a recommended practice and generally would just make engineers to not have trust in the type system. If you are facing a challenge typing something it is better to use some of the escape hatches documented below instead of just ignoring TS errors.
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,73 @@
...
+
+### 2. Favor efficiency instead of strict typing
+
+In ideal world all engineers would be capable of typing everything properly from the get go. However we know that is not the case and engineers needs time and practice to get there. For that reason we favor efficiency over strict typing. 
+
+Our philosophy is that over time engineers will organically get more familiar with TypeScript and stricness will come as a result. Code reviews are also a place where engineers might get quick feedback that would allow them to more efficiently adress typing issues instead of sinking a lot of time during initial development trying to figure it out on their own.
+
+When facing such typing challenges we recommend the following strategies as a escape hatch:
```
pattonwebz:
```suggestion
When facing such typing challenges we recommend the following strategies as an escape hatch:
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,180 @@
...
+## Recommended `tsconfig.json` settings
+
+This is not a full list but we highlight some recommended settings here.
+
+1. `noFallthroughCasesInSwitch: true`- [Definition](https://www.typescriptlang.org/tsconfig/#noFallthroughCasesInSwitch) - ensures you are not skipping anything in a switch statement, and always at least provide a default option.
+2. `noImplicitAny: true` - [Definition](https://www.typescriptlang.org/tsconfig/#noImplicitAny) - ensures you define `any` explicitly, rather than implicitly assume it. This makes it a lot easier to find all code relying on any for future refactoring.
+3. `strictNullChecks: true` - [Definition](https://www.typescriptlang.org/tsconfig/#strictNullChecks) - forces handling both `undefined` and `null`
+4. `noUnusedParameters: true` - [Definition](https://www.typescriptlang.org/tsconfig/#noUnusedParameters) - does not allow unused parameters to remain in code
+
+## TypeScript and React
```
tobeycodes, line 96: **question:** are all of these in the eslint-config? If no, I think we should add them. Preferences like this are hard to remember

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,180 @@
...
+### Use `type` for React Props
+
+While `interface` and `type` are mostly interchangeable, prefer `type` over `interfaces` unless you expect the props to be the base for other components's props.
+
+```ts
+type BlocksProps = {
+	html: string;
+};
+
+export const Blocks = ({ html }: BlocksProps) => {
```
tobeycodes: **question:** I know you're not a fan of FC, but do you think we should encourage adding return types for these examples? I think return types are important (even for React components) as sometimes they can return different things like string, undefined, etc. I think it'll help build a happen to write them for non React component functions too

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,180 @@
...
+
+### 2. Favor efficiency instead of strict typing
+In an ideal world, all engineers would be capable of typing everything properly from the get-go. However, we know that is not the case and engineers need time and practice to get there. For that reason, we favor efficiency over strict typing. 
+
+Our philosophy is that over time engineers will organically get more familiar with TypeScript and strictness will come as a result. Code reviews are also a place where engineers might get quick feedback that would allow them to more efficiently address typing issues instead of wasting a lot of time during initial development trying to figure it out on their own.
+
+When facing such typing challenges we recommend the following strategies as an escape hatch:
+#### 1. The `as unknown as ExpectedType` type cast
```
tobeycodes, line 45: **nit:** no empty line between the paragraph and heading. I think prettier can auto-fix this

_includes/markdown/Typescript.md
tobeycodes, line 1: Great document. Definitely like these escape hatches

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,180 @@
...
+
+While we'd like none of these escape hatches to be used, it would be unrealistic to expect all engineers to have types 100% correct on every project all the time. We believe that these recommendations provide a path to address them organically during the life-cycle of the project.
+
+## TypeScript and CI
+
+We recommend that all projects run the typescript compiler to check type safety at the CI level. Every Pull Request/Merge Request should run `tsc --noEmit` on CI to ensure there are no compilation issues before merging the branch.
+
+This is a great strategy to avoid builds failing due to type errors. With Next.js for instance, you might not realize you have a TS compilation issue until a build is executed, therefore to avoid merging code that produces failing builds we recommend always ensuring that the code has no TypeScript compilation issues before merging.
+
+## Recommended `tsconfig.json` settings
```
tobeycodes, line 85: **thought:** could we link to the tsconfig.json here that is recommended on projects?

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,180 @@
...
+const Layout = (props: LayoutProps) {
+    return (
+        <main>
+            {children}
+        </main>
+    );
+}
+```
+
+### Avoid prop spreading
```
tobeycodes, line 166: ❤️

tobeycodes (approved):

_includes/markdown/Typescript.md
nicholasio: yes they are!

_includes/markdown/Typescript.md
nicholasio: Wouldn't `string`, `undefined` be a valid `JSX.Element`? so does it really matter? I'm not necessarily against it just haven't found it to be very useful. But I'm open to it if it makes sense.

@kasparszarinovs any thoughts here?

_includes/markdown/Typescript.md
nicholasio: done

_includes/markdown/Typescript.md
nicholasio: fixed

_includes/markdown/Typescript.md
tobeycodes: They would not. Things like `string`, `null`, `undefined` would not be valid JSX.Element. https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L327

My personal preference is to always add return types over having TS infer them but I'm not that bothered by it either if we don't

tlovett1 (approved):

_includes/markdown/Typescript.md
```diff
@@ -19,7 +19,7 @@ The main reason we don't require TypeScript for standard WordPress projects at t
 
 ## How to use TypeScript
 
-In an ideal world, every engineer would be fully capable of writing all sorts of complex TypeScript types, however, we understand that it takes time to master a language like TypeScript, especially in a few scenarios that require writing more complex types. Therefore, while we use a reasonably strict `tsconfig.json`` we are flexible in practice. 
+In an ideal world, every engineer would be fully capable of writing all sorts of complex TypeScript types, however, we understand that it takes time to master a language like TypeScript, especially in a few scenarios that require writing more complex types. Therefore, while we use a reasonably strict `tsconfig.json` we are flexible in practice.
```
kasparszarinovs: @nicholasio I would argue that this:
```
every engineer would be fully capable of writing all sorts of complex TypeScript types
```
is something you'd ideally want to avoid 😅 While TypeScript can be absolutely amazing as to what it can achieve (see Prisma dynamically generated DB types for example) - we feel like this adds an immense complexity overhead that is not needed. It let's you do things that no statically typed language lets you do - and this can add a lot of complexity and confusion to code. Obviously, there are times where it's super helpful, but in most cases it's likely to take up too much time for something that could be done simpler - and this is what, I think, you should try and promote - using TS, but not going overboard with it. Basically - use it like you would with Java, .NEt or similar types.

I'm well aware I'm not providing an alternative for the particular highlighted text, instead I am trying to provide some insight as to how far you may want to go with TS, and at what point would you may want to stop 😅

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,188 @@
+The purpose of this document is to help guide Engineers in the usage of TypeScript at 10up projects. The goal here is not to teach TypeScript but rather explain the 10up stance on when to use TypeScript alongside general recommendations.
+
+## Why TypeScript
+
+We believe that:
+1. TypeScript can make engineers more confident
+2. TypeScript makes refactoring code easier
+3. TypeScript helps engineers unfamiliar with the codebase
```
kasparszarinovs, line 8: There's also a case for TS helping to catch certain issues earlier.

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,188 @@
...
+
+The main reason we don't require TypeScript for standard WordPress projects at the moment is that TypeScript support for block development is still very limited, largely driven by community types packages that don't always match the actual types for the `@wordpress` packages.
+
+## How to use TypeScript
+
+In an ideal world, every engineer would be fully capable of writing all sorts of complex TypeScript types, however, we understand that it takes time to master a language like TypeScript, especially in a few scenarios that require writing more complex types. Therefore, while we use a reasonably strict `tsconfig.json` we are flexible in practice.
+
+We want the TypeScript to always be valid and issue no errors at compilation, but we allow a few escape hatches should engineers get stuck with a particular typing problem. We follow the principles below when writing TypeScript:
+
+### 1. Never ship code that has compilation issues
```
kasparszarinovs, line 31: @nicholasio additionally, you could provide a standard CI action for this to run `tsc --noEmit` - this could prevent PRs with issues being merged.

[EDIT] Nvm, noticed you mention this further down.

_includes/markdown/Typescript.md
tobeycodes: @kasparszarinovs Do you have an example of going overboard with TS? It is not something I feel like I have ran into on large and complex TS projects. So for my benefit I'd love to understand more about this and perhaps I'm going overboard and not realising it and then making things more complicated for other engineers

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,188 @@
...
+
+While this usually indicates that there's something wrong with types or even the code itself, this would at least provide type checking for the subsequent lines of code. This strategy, for instance, is much better than just typing as any.
+
+#### 2. The `any` type.
+
+Another strategy is to just type your variable with `any` which essentially disables type checking for that variable and while it is a more drastic measure it should be an easy way to unblock work until engineers have more time to come back and try to get the types right.
+
+Note: we recommend setting the `noImplicitAny` setting to `true` in `tsconfig.json` which forces explicitly setting a type as any. This makes it easier to identify code that is relying on any and would also make catching the escape hatch in code reviews which is a great opportunity to get feedback and potentially get the types right.
+
+#### 3. The `// @ts-expect-error` comment
```
kasparszarinovs, line 62: Ideally this would be avoided if possible - casting as `any` should nearly always fix this issue. If you're having to use this - it may be a case where the engineer could ask for someone more experienced with TS to assist with the issue. Would also work as a great learning experience.

_includes/markdown/Typescript.md
kasparszarinovs: I would agree with @tobeycodes  that return types are important. We actually tend to enforce them on the Node side (not yet on FE) - this helps ensure that your code works as expected.

I may be a bit out of the loop as to why `React.FC` is not favoured - so please do let me know what I'm missing. Personally, however, I don't see n issue with using those.

tl;dr; - yes, do use return types if you can :)

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,188 @@
...
+}
+```
+
+### Use function default arguments instead of `defaultProps`
+
+`defaultProps` has been deprecated in React 19, therefore prefer using function default arguments.
+
+```ts
+type MyComponentProps = {
+    title?: strinmg;
```
kasparszarinovs:
```suggestion
    title?: string;
```

_includes/markdown/Typescript.md
kasparszarinovs: And another ❤️

_includes/markdown/Typescript.md
kasparszarinovs: @tobeycodes I do, but I'll have to manually re-create the situation as we normally reason them as part of PR process 😅 I'll send something over during next week if that works!

_includes/markdown/Typescript.md
tobeycodes: Another thing perhaps worth adding, if you're using NPM packages in TS then upgrades dependencies become less risky. If you're using their types and their types change or you're using values that no longer exist your builds will fail for TS errors. I think it makes easier to use something like dependabot and merge package updates more often

_includes/markdown/Typescript.md
nicholasio: I've used to use React.FC a lot but a few things happened:

1. Until React 17 (I think) it would always include `children` even if you didn't need it.
2. Since React 18 it does not include `children` anymore
3. Most of the times we don't need the other things from React.FC so it's a littler easier for beginners not have to think about what React.FC is
4. We likely have projects still using React < 18, which could potentially run into a confusion if deciding to use TS before upgrading React (and its types packages).

Since React 18 and recent versions of TS though there's [nothing wrong with it](https://www.totaltypescript.com/you-can-stop-hating-react-fc) (altought some still prefer just typing props), I just feel like it could be confusing given its history and projects on older versions of react.

However between typing return types and just using React.FC I can see React.FC being better, perhaps we just need to add a note about it. In any case, I feel like we would need a new section since I'm not even mentioning either one in the doc today :)

Regarding return types I know it's a hot debate - I personally haven't found issues with most functions/components. Of course in some cases you should definately add a return type. So I tend to only type when needed.

nicholasio: I'll draft up a new section around `Typing React Components` and `Return Types`. If you have any suggestions/wording feel free to drop a comment here.

_includes/markdown/Typescript.md
nicholasio: @kasparszarinovs I see where you're coming from, definately not the intention I'll rewrite this to better convey the message that "ideally no one would struggle with TS"

_includes/markdown/Typescript.md
tobeycodes: I personally think this is a great example of where React core team did something, then realized it was a bad decision and then made a breaking change to fix it. I personally think we should make decisions based on the latest release of a framework. The beauty of TS here is that users are going to know immediately whether their FC type is correct whether they use children or not because of the feedback they get in their IDE or at build time. 

If projects have the freedom they should use TS first before upgrading React. That is going to help highlight any breaking changes automatically when they upgrade and FC would be one example of this where types need to change (which can be completely automated with codemods)

I also think using FC is great for beginners because they are able to do a Find by Reference and see what else a function component can do.

My 2 cents on return types is the more engineers do it the faster they will learn TS, but also we shouldn't force it

But at the end of the day, I'm not really that opinionated on all of this. The path of least resistance for most engineers is best

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,188 @@
...
+2. TypeScript makes refactoring code easier
+3. TypeScript helps engineers unfamiliar with the codebase
+
+Therefore we are big fans of TypeScript at 10up and we strive to use it whenever it makes sense.
+
+## When to use TypeScript
+
+At 10up we require that all projects considered "JavaScript-First" to make use of TypeScript. Generally speaking Standard WordPress projects (including block development) would not fall under "JavaScript-First" projects. 
+
+If you are building a project using Next.js and/or 10up's [HeadstartWP framework](https://headstartwp.10up.com/), Sanity and any other headless CMS'es powered by a React front-end then we require TypeScript unless other circumstances make using just JavaScript a better choice.
```
dainemawer: @nicholasio remove "just" here - `make using JavaScript a better choice"

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,188 @@
...
+
+The main reason we don't require TypeScript for standard WordPress projects at the moment is that TypeScript support for block development is still very limited, largely driven by community types packages that don't always match the actual types for the `@wordpress` packages.
```
dainemawer: @nicholasio remove "at the moment" - `for standard WordPress projects is that TypeScript support...`

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,188 @@
+The purpose of this document is to help guide Engineers in the usage of TypeScript at 10up projects. The goal here is not to teach TypeScript but rather explain the 10up stance on when to use TypeScript alongside general recommendations.
+
+## Why TypeScript
```
dainemawer, line 3: It might be worth mentioning that many industry standard tools and packages (Sanity, NextJS, Storybook etc) have moved to TypeScript as well - its not something we're considering in isolation, its a noticeable industry pattern that's shifting development for the better

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,188 @@
...
+
+We believe that:
```
dainemawer, line 5: Theres also something to be said about education here - and sharpening engineers skills to be more efficient and produce code with less bugs

_includes/markdown/Typescript.md
dainemawer: I think the issue here is that implementing TypeScript in certain circumstances can force engineers down a rabbit hole, especially when combining intricate types across technologies or implementations - for instance, `ReactElement` and the `HTMLButtonElement`. 

I've experienced this myself on projects where getting types right becomes a bottleneck, and the value begins to slip. Unfortunately, I think this essentially boils down to experience—these situations aren't frequent—they're more like edge cases, which can be solved mainly by reaching out to more experienced team members or looking at code. This should be communicated to team members who need more experience in the language.

That said, like with most technologies, we should anticipate a slow ramp up for most engineers unfamiliar with TypeScript and its conventions

_includes/markdown/Typescript.md
dainemawer: @nicholasio - most frameworks wont even trigger a successful build if there's a TypeScript error - surely the CI would catch this if running a build script? Theres no way (as far as I know) other than using an `@ts-ignore` to get around this?

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,188 @@
...
+
+In an ideal world, every engineer would be fully capable of writing all sorts of complex TypeScript types, however, we understand that it takes time to master a language like TypeScript, especially in a few scenarios that require writing more complex types. Therefore, while we use a reasonably strict `tsconfig.json` we are flexible in practice.
+
+We want the TypeScript to always be valid and issue no errors at compilation, but we allow a few escape hatches should engineers get stuck with a particular typing problem. We follow the principles below when writing TypeScript:
+
+### 1. Never ship code that has compilation issues
+
+Depending on the tooling you're using (e.g. Next.js) it is possible to disable type checking and still compile invalid TS code as long as the underlying JS code is valid. While the runtime behavior might still be working as expected this is not a recommended practice and generally would just make engineers not have trust in the type system. If you are facing a challenge typing something it is better to use some of the escape hatches documented below instead of just ignoring TS errors.
+
+### 2. Favor efficiency instead of strict typing
```
dainemawer, line 35: @nicholasio, we need to define "efficiency" here - it's a bit confusing in this context. Are we saying that if you cannot figure out a type error its better to use and escape hatch? Perhaps mentioning something around time-boxing issues is more prescriptive.

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,188 @@
...
+In an ideal world, all engineers would be capable of typing everything properly from the get-go. However, we know that is not the case and engineers need time and practice to get there. For that reason, we favor efficiency over strict typing. 
+
+Our philosophy is that over time engineers will organically get more familiar with TypeScript and strictness will come as a result. Code reviews are also a place where engineers might get quick feedback that would allow them to more efficiently address typing issues instead of wasting a lot of time during initial development trying to figure it out on their own.
```
dainemawer, line 41: @nicholasio is there a viable type-checker we could add to CI that would flag potential TypeScript errors?

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,188 @@
...
+
+When facing such typing challenges we recommend the following strategies as an escape hatch:
+
+#### 1. The `as unknown as ExpectedType` type cast
+
+When struggling with a type that does not match the expected type, one of the escape hatches we recommend is to forcibly typecast the type to the expected type:
+
+```ts
```
dainemawer, line 49: @nicholasio, if these escape hatches are used, I think it would make sense as part of this guidance to instruct engineers to leave a note or `@todo` as to what issue they are experiencing. We do something similar in CSS, leaving an inline comment next to `!important` to explain the intention of its use. 

As much as we want to maintain efficiency here, these escape hatches can cause a significant amount of technical debt if we're not careful, so we need to be conscious of that.

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,188 @@
...
+While we'd like none of these escape hatches to be used, it would be unrealistic to expect all engineers to have types 100% correct on every project all the time. We believe that these recommendations provide a path to address them organically during the life-cycle of the project.
+
+## Tools
+
+- [TypeScript Error Translator](https://ts-error-translator.vercel.app/): Lets you drop any TypeScript error message and gives you a human-readable explanation in plain English.
+- [pretty-ts-errors] (https://marketplace.visualstudio.com/items?itemName=yoavbls.pretty-ts-errors): A VSCode extension that translaters TypeScript errors in a human-readable way right into VSCode.
+
+## TypeScript and CI
+
+We recommend that all projects run the typescript compiler to check type safety at the CI level. Every Pull Request/Merge Request should run `tsc --noEmit` on CI to ensure there are no compilation issues before merging the branch.
```
dainemawer, line 81: @nicholasio perhaps we should also run this in `husky` / `lint-staged` before code is committed?

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,188 @@
...
+}
+```
+
+### Use function default arguments instead of `defaultProps`
+
+`defaultProps` has been deprecated in React 19, therefore prefer using function default arguments.
+
+```ts
+type MyComponentProps = {
+    title?: string;
```
dainemawer, line 137: @nicholasio the optional operator here has always bugged me - the semantics makes it difficult to really determine what props are actually required for a component to function in a fundamental way. Is there guidance around this? When should a type leverage an optional prop vs an optional prop with a default? The two are not exclusive and can lead to overuse which makes components brittle

_includes/markdown/Typescript.md
dainemawer: On some projects we've used this convention:

```
const Header: React.FC<ComponentProps> = () => {}
```

It was a nice convention, Im not sure it holds any value of any other approach but I found it more readable in general.

_includes/markdown/Typescript.md
nicholasio: Next.js allows for disabling TS checks at build time so in theory it's possible.  I think running full builds would be at the discretion of each project. The main message I'm trying to convery is that typecheck should be executed at the PR level.

_includes/markdown/Typescript.md
nicholasio: `tsc --noEmit`

_includes/markdown/Typescript.md
nicholasio: Unfortunately `tsc --noEmit` on its own doesn't work with husky/lint-staged requiring other third-party tools.

I think the IDE support and CI checks will cover this, I didn't want to introduce another non-official TS CLI into the mix

_includes/markdown/Typescript.md
nicholasio: @dainemawer What you're talking about is the cost of ramping up engineers to TypeScript to which we hope to minimize with the tone and expectations that we're setting up here alongside the escape hatches we are strategically allowing.

The other side of the coin which we hope to address as well here is overengineering a codebase with types that are more complex than it needs to be for the limited purposes of projects.

_includes/markdown/Typescript.md
tobeycodes: @nicholasio That's not true. I am going to send you a link to an internal project on Slack that shows how to do it

nicholasio: @tobeycodes I was reffering to the fact that if you wanted to only do the tsc check for the files you are modifying it wouldn't work with `tsc` https://github.com/microsoft/TypeScript/issues/27379

But if you want to run TSC for the whole project then it should be fine.

_includes/markdown/Typescript.md
tobeycodes: I do not think this is a problem that is limited to TS. For example, you still have to decide the same thing when writing OOP PHP. 

In my opinion all keys of an interface should be required and no functions should have default values until you have a good reason to do either. Some examples I can think of
- If you have a React component that accepts a description field but in some cases you don't want to show a description that would be a good example of using an optional key in an interface. Then make sure the code has `{description ? <p>{description}<p/> : null}`
- If you have a Button component that accepts the type (submit, button, reset) then probably a good argument for type to have a default 

I agree with you that overuse can create problems which is why I think you need a good reason to use them. And it's important if you use them that you have some sort of way automating testing of the function or component when you supply all the keys and only the required keys (e.g. vrt if it's storybook, jest if it's functions, etc)

_includes/markdown/Typescript.md
tobeycodes: I ❤️ TODOs. I'm a big fan of engineers using them and also leaving their name with it so they take ownership of it `TODO(tobeycodes): Something...` which I think is important so they come back and fix it later. So often TODOs are left in the codebase forever. 

This is a great plugin for VSCode users to help manage them too https://marketplace.visualstudio.com/items?itemName=Gruntfuggly.todo-tree

_includes/markdown/Typescript.md
tobeycodes: Perhaps there should be an example of importing a type from one of these packages and using it in some example code. I think we should try to encourage engineers to import and use types from packages where possible rather than write their own (unless it will slow them down to do so)

_includes/markdown/Typescript.md
nicholasio: what if we use eslint to detect todos and show as warnings 🤔 that wound't prevent todos from being left in the codebase forever but would be an easy way to list them.

_includes/markdown/Typescript.md
tobeycodes: @dainemawer Perhaps this document could give some examples on how to solve common problems like React onClick event handling. That way engineers can check that list rather than going down the rabbit hole of Google (which will give you ton of mixed results) or asking another engineer

_includes/markdown/Typescript.md
nicholasio: Actually I think a better thing if we encourage todos is to have a eslint rule to enforce the `TODO(name)` pattern. I don't think it makes sense for the todo itself to be a warning.

_includes/markdown/Typescript.md
tobeycodes: In my experience the easiest way to assist with TS types for other engineers is once it's in a PR and I can see all the code. I've found Slack messages in isolation without seeing all the code hard to answer. So perhaps encourage them to get their work done with `any` and then TS types support comes once the PR is open. It also leaves a feedback trail for any other engineers on the project

_includes/markdown/Typescript.md
tobeycodes: I actually do not like warnings. I think they create a lot of noise and more confusion for engineers where they are not sure whether they need to make a change or if it is ok to leave their code as it is. I'd rather make it clear to engineers what needs to be fixed and otherwise leave their code. Similar to TODOs I don't think warnings ever get cleaned up.

https://www.npmjs.com/package/eslint-plugin-only-warn

_includes/markdown/Typescript.md
nicholasio: I think that would out of scope for this particular PR.

_includes/markdown/Typescript.md
nicholasio: @kasparszarinovs I agree but I also feel like we need this as a last resort. Like @tobeycodes said this could potentially be a easy fix during code review.

_includes/markdown/Typescript.md
nicholasio: I'll add this to the HeadstartWP scaffold, not sure if we need to mention it here.

_includes/markdown/Typescript.md
nicholasio: Added a paragraph about https://react-typescript-cheatsheet.netlify.app/ I think that will suffice for now. We can interate and document the things we see people struggling the most in the future.

_includes/markdown/Typescript.md
nicholasio: Done

_includes/markdown/Typescript.md
nicholasio: Added a note about the TODOs

_includes/markdown/Typescript.md
nicholasio: Also updated the wording here to avoid overengineering types.

tobeycodes: Never knew that cheat sheet existed, so great idea to add

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,221 @@
...
+
+We believe that:
+1. TypeScript can make engineers more confident
+2. TypeScript makes refactoring code easier
+3. TypeScript helps engineers unfamiliar with the codebase
+4. TypeScript helps catch certain issues earlier
+
+Therefore we are big fans of TypeScript at 10up and we strive to use it whenever it makes sense.
+
+We also recgonize that the industry has largely adopted TypeScript as the defacto standard for writing JavaScript apps.
```
Antonio-Laguna: [nit] typo

```suggestion
We also recognize that the industry has largely adopted TypeScript as the de facto standard for writing JavaScript apps.
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,221 @@
...
+
+The main reason we don't require TypeScript for standard WordPress projects is that TypeScript support for block development is still very limited, largely driven by community types packages that don't always match the actual types for the `@wordpress` packages.
+
+## How to use TypeScript
+
+In an ideal world, no one would struggle with TypeScript, however, we understand that it takes time to master a language like TypeScript, especially in a few scenarios that require writing more complex types. Therefore, while we use a reasonably strict `tsconfig.json` we are flexible in practice.
+
+We also don't want engineers looking for "perfection" when it comes to writing complex types. TypeScript has a very powerfull type system and it allows for amazing things to be done statically at compilation level, however, we should recognize that most projects will suffice with simpler, easy-to-use types. Unless there's a good reason (e.g building a reusable package or library), engineers should strive to not write overly complex types.
```
Antonio-Laguna: [nit] typos

```suggestion
We also don't want engineers looking for "perfection" when it comes to writing complex types. TypeScript has a very powerful type system and it allows for amazing things to be done statically at the compilation level, however, we should recognize that most projects will suffice with simpler, easy-to-use types. Unless there's a good reason (e.g. building a reusable package or library), engineers should strive to not write overly complex types.
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,221 @@
...
+
+### 1. Never ship code that has compilation issues
+
+Depending on the tooling you're using (e.g. Next.js) it is possible to disable type checking and still compile invalid TS code as long as the underlying JS code is valid. While the runtime behavior might still be working as expected this is not a recommended practice and generally would just make engineers not have trust in the type system. If you are facing a challenge typing something it is better to use some of the escape hatches documented below instead of just ignoring TS errors.
+
+### 2. Favor efficiency instead of strict typing
+
+In an ideal world, all engineers would be capable of typing everything properly from the get-go. However, we know that is not the case and engineers need time and practice to get there. For that reason, we favor efficiency over strict typing. 
+
+By "efficiency" we mean engineers recognizing and time-boxing the time spent trying to type something properly and not letting that block the development of the task at hand. We prefer to ship non-strict typed code instead of sinking a lot of time on it.
```
Antonio-Laguna: [nit] I feel this comes from "time sink" but I don't think it is the right verb.

```suggestion
By "efficiency" we mean engineers recognizing and time-boxing the time spent trying to type something properly and not letting that block the development of the task at hand. We prefer to ship non-strict typed code instead of spending a lot of time on it.
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,221 @@
...
+	/* */
+};
+
+export const MyComponent: React.FC<MyComponentProps> = (props) => {
+	// component logic
+}
+```
+**Note**: If using React < 17, be aware that `children` will be explicitly set for all components using `React.FC`. This can cause some TS errors when upgrading React to 18+, if a component is passed `children` but does not explicitly declare it as a prop.
+
+The [React TypeScript Cheatsheet](https://react-typescript-cheatsheet.netlify.app/) is a good resource to check how to type specific things in React (e.g Forms, Events etc).
```
Antonio-Laguna: [nit] missing dot

```suggestion
The [React TypeScript Cheatsheet](https://react-typescript-cheatsheet.netlify.app/) is a good resource to check how to type specific things in React (e.g. Forms, Events etc).
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,221 @@
...
+type MyComponentProps = {
+    title?: string;
+};
+
+export const MyComponent = ({ title = 'Default Title' }: MyComponentsProps) => {
+    // ...
+}
+```
+
+Make sure the eslint rule `react/require-default-props` is set up to look for `defaultArguments` e.g `'react/require-default-props': ['error', { functions: 'defaultArguments' }]`;
```
Antonio-Laguna:
```suggestion
Make sure the ESLint rule `react/require-default-props` is set up to look for `defaultArguments` e.g `'react/require-default-props': ['error', { functions: 'defaultArguments' }]`;
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,221 @@
...
+
+For the most part, the React Component types should be colocated with the React component, meaning that it should be in the same file that the React component is written. Only hoist types to avoid circular dependencies and when you expect them to be reused/shared across many components/files.
+
+### Use `@types` packages if necessary
+
+If the library you're using does not ship types, check if there is type information available for that package in the [DefinatelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) repo.
+
+### Global types definitions
+
+**Do not** put the application types in the global scope just to avoid importing them. You may only create global type definitions when needing to extend global types (e.g `window` object).
```
Antonio-Laguna:
```suggestion
**Do not** put the application types in the global scope just to avoid importing them. You may only create global type definitions when needing to extend global types (e.g. `window` object).
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,221 @@
...
+
+Therefore we are big fans of TypeScript at 10up and we strive to use it whenever it makes sense.
+
+We also recgonize that the industry has largely adopted TypeScript as the defacto standard for writing JavaScript apps.
+
+## When to use TypeScript
+
+At 10up we require that all projects considered "JavaScript-First" to make use of TypeScript. Generally speaking Standard WordPress projects (including block development) would not fall under "JavaScript-First" projects. 
+
+If you are building a project using Next.js and/or 10up's [HeadstartWP framework](https://headstartwp.10up.com/), Sanity and any other headless CMS'es powered by a React front-end then we require TypeScript unless other circumstances make using JavaScript a better choice.
```
Antonio-Laguna:
```suggestion
If you are building a project using Next.js and/or 10up's [HeadstartWP framework](https://headstartwp.10up.com/), Sanity and any other headless CMSs' powered by a React front-end then we require TypeScript unless other circumstances make using JavaScript a better choice.
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,221 @@
...
+    // render blocks
+}
+```
+
+### Use `React.FC` to type React Components
+
+When typing React components prefer using `React.FC`. While in the past it was problematic because of implicitly accepting `children` as an optional prop, since React 18 it [no longer does that](https://www.totaltypescript.com/you-can-stop-hating-react-fc). Using React.FC ensures your component is correctly typed, including its return type.
+
+```tsx
+import * as React from 'react';
```
Antonio-Laguna, line 117: Wasn't this discouraged since React 17? I've been using 

```
import { FC } from 'react';
```

or

```
import type { FC } from 'react';
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,221 @@
...
+
+type MyComponentProps = {
+	/* */
+};
+
+export const MyComponent: React.FC<MyComponentProps> = (props) => {
+	// component logic
+}
+```
+**Note**: If using React < 17, be aware that `children` will be explicitly set for all components using `React.FC`. This can cause some TS errors when upgrading React to 18+, if a component is passed `children` but does not explicitly declare it as a prop.
```
Antonio-Laguna, line 127: While this is true, I'd say it's been a while since React 18 is out (2 years!) While it doesn't harm I don't feel this is as relevant now? Just flagging to avoid this feeling dated when it's born but feel free to disregard

Antonio-Laguna (approved):
@nicholasio great stuff! I just stumbled upon this and thought I'd give it a read! Found a few minor typos and left a comment for your consideration.

Great work!

_includes/markdown/Typescript.md
nicholasio: I believe It was `import React from 'react'` that was discouraged.

_includes/markdown/Typescript.md
nicholasio: The only reason I added this was bc there might be projects still on React 17.

_includes/markdown/Typescript.md
dainemawer: Agreed, thanks for the insight @nicholasio @tobeycodes - I think some top of mind solutions to challenges can even come after this Best Practice doc is released. Fabian and I started a Git Discussions for FSE which could be helpful for TypeScript as well: https://github.com/10up/gutenberg-best-practices/discussions - the team has already started logging questions there for Gutenberg, may be a nice solution for TypeScript.

_includes/markdown/Typescript.md
dainemawer: Great points @tobeycodes 

@nicholasio agreed, though it would be good to get guidance on this at some point.

_includes/markdown/Typescript.md
dainemawer: Good to know - it would be great if we could get the linter to pick that up but maybe that will come with time.

_includes/markdown/Typescript.md
dainemawer: Great, thanks @nicholasio @tobeycodes - I like the idea of eslint handling this.

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,221 @@
...
+    title: string,
+    category: string,
+    link: string
+    // other props
+};
+
+const Post = (props: PostProps) => {
+    return (
+        <article>
+            <PostContent {...props}>
```
aaemnnosttv:
```suggestion
            <PostContent {...props} />
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,221 @@
...
+
+`PropsWithChildren` will make `children` optional, but sometimes you do want to be very explicit and make it required. Therefore we generally recommend declaring `children` explicitly.
+
+```tsx
+type LayoutProps = {
+    children: React.Node;
+};
+
+// If Layout is not passed a children, TS will catch the error
+const Layout = (props: LayoutProps) {
```
aaemnnosttv:
```suggestion
const Layout = (props: LayoutProps) => {
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,221 @@
...
+To extend global types create a `global.d.ts` file and include them in your `tsconfig.json` in the `include` config option. Here's an example for a Next.js project:
+
+```json
+ "include": [
+    "next-env.d.ts",
+    "src/global.d.ts",
+    "**/*.ts",
+    "**/*.tsx"
+  ],
+```
```
aaemnnosttv:
```suggestion
```json
{
  "include": [
    "next-env.d.ts",
    "src/global.d.ts",
    "**/*.ts",
    "**/*.tsx"
  ],
}
```
```

File: _includes/markdown/Typescript.md [hunk 40]
```diff
@@ -0,0 +1,221 @@
...
+
+## Why TypeScript
+
+We believe that:
+1. TypeScript can make engineers more confident
+2. TypeScript makes refactoring code easier
+3. TypeScript helps engineers unfamiliar with the codebase
+4. TypeScript helps catch certain issues earlier
+
+Therefore we are big fans of TypeScript at 10up and we strive to use it whenever it makes sense.
```
aaemnnosttv:
```suggestion
Therefore, we are big fans of TypeScript at 10up and we strive to use it whenever it makes sense.
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,221 @@
...
+```js
+// TODO(nicholaiso): the types coming from the package do not seem correct
+```
+
+While we'd like none of these escape hatches to be used, it would be unrealistic to expect all engineers to have types 100% correct on every project all the time. We believe that these recommendations provide a path to address them organically during the life-cycle of the project.
+
+## Tools
+
+- [TypeScript Error Translator](https://ts-error-translator.vercel.app/): Lets you drop any TypeScript error message and gives you a human-readable explanation in plain English.
+- [pretty-ts-errors](https://marketplace.visualstudio.com/items?itemName=yoavbls.pretty-ts-errors): A VSCode extension that translaters TypeScript errors in a human-readable way right into VSCode.
```
aaemnnosttv:
```suggestion
- [pretty-ts-errors](https://marketplace.visualstudio.com/items?itemName=yoavbls.pretty-ts-errors): A VSCode extension that translates TypeScript errors in a human-readable way right into VSCode.
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,221 @@
...
+3. `strictNullChecks: true` - [Definition](https://www.typescriptlang.org/tsconfig/#strictNullChecks) - forces handling both `undefined` and `null`
+4. `noUnusedParameters: true` - [Definition](https://www.typescriptlang.org/tsconfig/#noUnusedParameters) - does not allow unused parameters to remain in code
+
+We also have a recommended [tsconfig.json](https://github.com/10up/headstartwp/blob/develop/projects/wp-nextjs/tsconfig.json) for Next.js projects.
+
+## TypeScript and React
+
+### Use `type` for React Props
+
+While `interface` and `type` are mostly interchangeable, prefer `type` over `interfaces` unless you expect the props to be the base for other components's props.
```
aaemnnosttv:
```suggestion
While `interface` and `type` are mostly interchangeable, prefer `type` over `interfaces` unless you expect the props to be the base for other components' props.
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,221 @@
...
+        </article>
+    );
+}
+```
+
+While there might be use cases where you need to use them, it's generally better to avoid them and pass props explicitly. In the example above it would be better to just pass a `post` object as a single prop.
+
+Prop spreading is especially dangerous when prop spreading into native DOM elements. So make sure to not ignore TypeScript errors and make sure your types are set up correctly so that TS can catch forwarded props incompatible with the target component/element.
+
+### Colocate types with React components
```
aaemnnosttv:
```suggestion
### Co-locate types with React components
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,221 @@
...
+
+For the most part, the React Component types should be colocated with the React component, meaning that it should be in the same file that the React component is written. Only hoist types to avoid circular dependencies and when you expect them to be reused/shared across many components/files.
```
aaemnnosttv:
```suggestion
For the most part, the React Component types should be co-located with the React component, meaning that it should be in the same file that the React component is written. Only hoist types to avoid circular dependencies and when you expect them to be reused/shared across many components/files.
```

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,221 @@
...
+
+### Use `@types` packages if necessary
+
+If the library you're using does not ship types, check if there is type information available for that package in the [DefinatelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) repo.
```
aaemnnosttv:
```suggestion
If the library you're using does not ship types, check if there is type information available for that package in the [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) repo.
```

aaemnnosttv (commented):
Thanks @nicholasio! I found a few more easy fixes with suggestions.

_includes/markdown/Typescript.md
kasparszarinovs: Keep in mind that `import type` is a TS only feature, and is stripped post compilation. It can be helpful, if you want to be extra sure the import only used as a type, but not necessary.

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,222 @@
...
+When struggling with a type that does not match the expected type, one of the escape hatches we recommend is to forcibly typecast the type to the expected type:
+
+```ts
+// force the variable to have the type `ExpectedType`
+const withExpectedType = myValueThatWithWrongType as unknown as ExpectedType;
+```
+
+While this usually indicates that there's something wrong with types or even the code itself, this would at least provide type checking for the subsequent lines of code. This strategy, for instance, is much better than just typing as any.
+
+#### 2. The `any` type.
```
joesnellpdx, line 56: @nicholasio 

Not critical and guidance is assumed, but for consistency - perhaps offer a working example of each escape hatch like you do for `as unknown as ExpectedType`

_includes/markdown/Typescript.md
```diff
@@ -0,0 +1,222 @@
...
+
+`PropsWithChildren` will make `children` optional, but sometimes you do want to be very explicit and make it required. Therefore we generally recommend declaring `children` explicitly.
+
+```tsx
+type LayoutProps = {
+    children: React.Node;
+};
+
+// If Layout is not passed a children, TS will catch the error
+const Layout = (props: LayoutProps) => {
```
devinle, line 157: @nicholasio should this be rewritten using React.FC as recommended above?

nicholasio: Made a quick follow-up commit to update this thanks

_includes/markdown/Typescript.md
nicholasio: Added thanks!

